Issue #2, opened 2013-05-17
 Uncaught TypeError: Cannot call method 'match' of undefined


brandongalbraith:
When upload is attempted by Chrome RECAP extension from PACER, the following error occurs:

Uncaught TypeError: Cannot call method 'match' of undefined

at line 128 in recap.js

audiodude:
I also see this error, looking into it now.

audiodude:
It appears the server is giving back the following for my upload/ call:

Content-Type:text/html; charset=utf-8
upload: pdf ignored.

This is a string, not a JSON object, so the parsing of it fails and then the response is just a string not parsed data, so it has no "message" property, so the code:

```
cb(object && object.message.match(/pdf uploaded/i));
```

raises the exception in the bug title.

mlissner:
Makes sense, and sounds like we can just catch the extension and ignore it, unless a fix on the server is possible.

PR #8, opened 2015-03-25
Ignore invalid JSON in responses to requests for it.

audiodude:
The real problem that causes #2 is that the server is not returning JSON to a JSON request. Presumably because the server considers it an error.

This PR attempts to fix this by just aborting the ready state change function and not calling the callback. It is assumed that the caller was expecting a parsed JSON data structure and since we are unable to provide that, the callback should be skipped.

To be clear, this simply sidesteps the issue and avoids the stack trace in the console. We still need to investigate why the server is returning such a response and if there is a more appropriate action the extension can take.

### DIFF
--- a/utils.js
+++ b/utils.js
@@ -111,7 +111,10 @@ function httpRequest(url, postData, responseType, callback) {
         if (responseType === 'json') {
           try {
             result = JSON.parse(result);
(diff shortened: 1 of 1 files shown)


mlissner:
Nice. Thanks.

PR #10, opened 2015-03-29
Fix donate: use click handler to add new tab when donate link is clicked

audiodude:
It looks like normal <a> links don't work in Chrome addon button panes. The link is suppressed somehow and doesn't open anywhere.

This CL makes the link functional by "manually" opening a tab when the link is clicked. The `href` property of the original link is used as the tab destination, so someone editing the HTML doesn't have to think too hard about this process.


mlissner:
Nice. Thanks. Serves me right to assume a link would work...

PR #11, opened 2015-03-31
First round of Jasmine tests

audiodude:
These are far from complete, but I thought I should commit them and create a Pull Request to make sure there is buy-in for the general methodology.

Note that the extra files like the Gruntfile and the contents of the `spec` folder, will get pulled into the built extension. However, as far as I can tell they have no effect, as expected.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+.DS_Store
+node_modules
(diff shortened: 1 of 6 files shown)


mlissner:
I've noticed how random junk seems to get pulled into the finished extension. Seems non-optimal, but I guess it's fine so far? 

Anyway, buy-in achieved. I don't know about Jasmine, have never touched Grunt, but I'm convinced you're going the right way.

mlissner:
Oh, and thanks for the docs too. I definitely will be needing them.

PR #13, opened 2015-04-03
Add remaining tests for the static PACER methods.

audiodude:
I added tests for the remaining methods. Also cleaned up some of the methods under test to return proper booleans, similar to what was done in the previous CL.

Thanks!


mlissner:
Love it.

PR #20, opened 2015-04-15
Remove nasty try/catch by using Jasmine spy in tests.

audiodude:
Simply add an empty object with a Jasmine spy, which can act as both a stub and a mock (it's acting as a stub here).


mlissner:
Lovely trick.

PR #21, opened 2015-05-09
Refactor content

audiodude:
This is a large complex change, even though it is largely a refactoring and there is supposed to be no changes to the functionality.

The purpose of the change is to make it so that content.js isn't just a top-level script that gets runs on each page that the extension attaches to. The way it was, it was basically untestable, because you had to run the whole script every time.

The refactoring introduces ContentDelegate which is a class that exposes methods that correspond to the methods that previously existed in content.js. Not all functions have been ported over to ContentDelegate in this pass, the more complex PDF uploading functionality will have to be done in a follow up.

To demonstrate the newfound testability of the new class, as well as to help verify that nothing has been broken, this PR also includes extensive Jasmine tests for the new class.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -2,9 +2,14 @@ module.exports = function(grunt) {
   grunt.initConfig({
     jasmine : {
       src : [
+        'assets/js/jquery.min.js',
+        'notifier.js',
         'pacer.js',
         'recap.js',
-        'utils.js'
+        'toolbar_button.js',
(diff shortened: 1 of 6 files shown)


mlissner:
Looks great and despite being a big refactor, I can't imagine how it could cause any issues. Sorry to keep you waiting.

audiodude:
Thanks! I should mention that I manually tested all of the functions that I changed.

PR #22, opened 2015-06-22
Refactor more of content.js and add tests

audiodude:
This PR refactors the part of content.js that is concerned with downloading PDF files, and adds tests for the same.

This part was left for its own PR because of the complexity of the PDF handling code. This code has several nested asynchronous functions. saves entire HTML documents in local variables, manipulates the state of the browser history, and asynchronously downloads then uploads PDF data.

Generally, the strategy employed was to separate the major functionality of this function and extract helper instance methods from several of the embedded functions that were defined inline. To this end, the functions ContentDelegate.showPdfPage and ContentDelegate.onDocumentViewSubmit were created. Liberal use was made of the "bind" browser function in order to assure that the proper object context was used when those functions are eventually called.

This PR was tested manually for both courts that return PDF and HTML documents.

### DIFF
--- a/content.js
+++ b/content.js
@@ -22,14 +22,16 @@ var pacer = importInstance(Pacer);
 var recap = importInstance(Recap);
 
 var url = window.location.href;
+var path = window.location.pathname;
 var court = PACER.getCourtFromUrl(url);
 var casenum = PACER.getCaseNumberFromUrl(document.referrer);
+var docid = PACER.getDocumentIdFromUrl(url);
 
 // Update the toolbar button with knowledge of whether the user is logged in.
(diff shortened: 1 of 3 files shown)


mlissner:
Did a brief skim and it looks like a huge improvement. I think I've got a couple changes, but I'm just going to merge here, pull `master` to my local repo, and then push my changes back up in a few minutes. 

Thanks for this. I think the "needs more tests" stage is done? If so that's a huge step forward.

audiodude:
Thanks, sounds good.

As far as needing more tests, the remaining code in content.js that *doesn't* call ContentDelegate is the part that still needs to be refactored and tested, but it's not that much code or work.

PR #23, opened 2015-06-25
Refactor some of the variables in the ContentDelegateSpec to be more sane

audiodude:
There were several places that were using the same or similar calls to new ContentDelegate. This change make it more sane that the work is all done in one place, which will make it easier if the signature of the ContentDelegate constructor changes.


mlissner:
Nice.

PR #24, opened 2015-06-25
Uncomment FormData tests and test them using mock/fake FormData objects

audiodude:
Based on getting more experience with Jasmine and how to mock built in types, this CL provides a fake FormData psuedo-class that can be used in the already written tests to assert against the values that the recap library is sending to the server.


mlissner:
Loving it. Jasmine is a pretty piece of kit.

PR #25, opened 2015-07-08
Create a gen204 call for logging purposes and its first usage.

audiodude:
The riskiest part of this change is the slight tweak to utils.js and the send method, so that if the formData is null it is not sent. I clicked around and things seemed to be working.

Otherwise, the gen204 method takes an Object that specifies parameters and does a GET request to the server's gen204 endpoint. I also added the specific event that we wish to log.

Thanks!


mlissner:
Looks good to me. I'll get it on the Chrome store momentarily.

audiodude:
Thank you!

PR #26, opened 2015-07-20
Refactor and test the handling of convertible document URLs and adding RECAP links

audiodude:
This is the final stage of the ambitious refactor of content.js. Now, instead of the content script being run and a bunch of naked functions and if statements being executed, the content script delegates to ContentDelegate, which means that the functionality of the content script is testable (and a sizable number of tests have been added).

In #25, we added a logging method to see if there were any so-called "convertible document URLs" that people were encountering in the wild. It turns out there's plenty of them. The RECAP chrome extension, when it encounters one of these, goes through a big dance to get the doc1/ style URL of the document, then finally uploads metadata for the document. Presumably this is to capture the docid, which might otherwise be unavailable if the document was actually viewed through the show_doc URL.

However, it seems to be the case that the PACER sites are re-writing the URLs of the show_doc style links so that on mouseover, they become doc1/ links. This would seem to make our efforts to pre-upload metadata for those documents less necessary.

However, in the interest of "leave this beast alone", I did exactly that and tested that after refactoring, the show_doc URLs triggered the proper uploadMetadata behavior.

Finally, the last if/for structures in content.js dealt with iterating over all of the links in the file and seeing if RECAP documents were available for any of them. This was, of course, tied up with the previous functionality, because they both operated over all the <A> tags. The solution was to make the links in the document a parameter to ContentDelegate. This dependency injection also proved useful when writing the tests.

I have tested that [R] icons properly show up on documents and that they function properly, both with and without the "show popups" preference.

Again, this is the final step of the refactoring and after it is merged, we can move on to bigger and better things.

### DIFF
--- a/content.js
+++ b/content.js
@@ -18,20 +18,20 @@
 
 var notifier = importInstance(Notifier);
 var toolbar_button = importInstance(ToolbarButton);
-var pacer = importInstance(Pacer);
-var recap = importInstance(Recap);
 
 var url = window.location.href;
 var path = window.location.pathname;
 var court = PACER.getCourtFromUrl(url);
 var casenum = PACER.getCaseNumberFromUrl(document.referrer);
 var docid = PACER.getDocumentIdFromUrl(url);
+var links = document.body.getElementsByTagName('a');
 
 // Update the toolbar button with knowledge of whether the user is logged in.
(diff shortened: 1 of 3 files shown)


mlissner:
So many lovely tests!

PR #27, opened 2017-11-03
Restore httpRequest() because jQuery is unreliable with binary files.

johnhawkinson:
See 3f1d2367169743fe4351c48a10e6890c4f7988bb: jQuery is very unreliable and breaks in mysterious unfathomable ways when dealing with binary data. Use XmlHttpRequest. Affects both Chrome and Firefox.

This was a showstopper bug.


johnhawkinson:
This corrects the showstopper (download corruption) failure, but is probably not ready for prime time. 
Looks like uploads aren't working right, among other things.

PR #29, opened 2017-11-16
Be polite. Every console.log() invokation should say "RECAP:"

johnhawkinson:
Also console.debug(), etc. Don't be one of those bad extensions where
you can't figure out who is sourcing the messages. Maybe there should
be a helper function, but that gets tricky because of the background
scope, so just do it the easy way for now.


mlissner:
Thanks. Man it's frustrating how helper functions aren't no-brainers...

PR #31, opened 2017-11-16
Options box design adjustment

johnhawkinson:
Address https://github.com/freelawproject/recap/issues/185
Give hierarchical parity in the options box and a bit more space for options.
Use the RECAP logo too.


options.html
```diff
@@ -8,40 +8,55 @@
...
+    </div>
+    <div>
+      <input type="radio" id="lawyer_style_filenames"
+             name="filename_style" value="lawyer_style_filenames">
+      <label for="lawyer_style_filenames">Use Lawyer style (e.g.
+        N.D.Cal._3&#8209;08&#8209;cv&#8209;03251_46_0.pdf)</label>
+    </div>
+  </fieldset>
+  
+  <!-- TBD: Optimize PACER -->
```
mlissner: Can you put a reference to the issue number in the comment here, or remove it completely? I don't know when we'll get to that ticket, so I want to make sure we don't have vestigial bits of it in here that we might not understand when we get around to it.

options.html
```diff
@@ -51,13 +66,15 @@
...
-    </p>
+  <p id="receipts_disabled" class="alert hidden"><i class="fa fa-warning gold"></i>&nbsp;Receipts are disabled in your PACER profile. While receipts are disabled, we are unable to upload PDFs that you purchase. To change this setting, please go to the billing preferences in your PACER account.
+  </p>
   </div>
   <div id="donate-plea" class="text-center">
-    <a href="https://free.law/donate/?referrer=recap-extension">
-      <img src="assets/images/donate-button.png">
+    <a href="https://free.law/donate/?referrer=recap-extension" style="text-decoration: none">
+      <img class="pad20" src="assets/images/donate-button.png">
+      <img class="pad20" width="150" src="assets/images/webstore-exports/440x280.png">
```
mlissner: I haven't looked at how this loads, but do we need a properly scaled down version of the RECAP logo? It's in the assets folder as an SVG that you can convert to whatever size. I'm happy to do so. I guess we need one that's 85px tall?

johnhawkinson:
> I haven't looked at how this loads, but do we need a properly scaled down version of the RECAP logo?

On what basis do we make such a decision? Are you asking me to profile the browser's rendering engine and assess the cost of CPU against the cost of space in the extension? No idea how to do either of those things.

Or are you asking if it looks pixelated and if that would change were it directly rendered from the SVG? I think that is a No.

mlissner:
OK, there's now a properly scaled version of the RECAP logo located at:

assets/images/recap-logo-143x85.png

Can you update to that, and the rest looks good to me.

johnhawkinson:
> Can you update to that, and the rest looks good to me.

Done, with a rebase to your master. Although it seems like this was a change you could have just amde yourself `:)`

mlissner:
Thanks for this, John. Good stuff.

PR #33, opened 2017-11-19
Display version number at bottom of options box.

johnhawkinson:
Fixes  freelawproject/recap#211

This is handy to have for checking the version without having to open about:addons.
It's most useful to developers (who else is worried about versions?)

Other extensions have visible version numbers in their dialogs, we're not breaking new ground.

Proof:
screen shot 2017-11-18 at 20 38 22


mlissner:
Looks about right to me. Thanks.

mlissner:
One thing - is there a reason we need to do the show/hide business? Why not have the div shown and just add the correct text? (I just pushed a tweak with that implemented.)

johnhawkinson:
You're correct, there's no need for the hidden -> unhidden transition.
You and your `let`s.

PR #34, opened 2017-12-08
Parsing improvements

johnhawkinson:
A variety of improvements that are on the way to appellate support, including parsing of caseids from URLs, adding some comments/questions, removing unnecessarily web-accessible js assets, relaxing the doc1 link regexp, etc.

### DIFF
--- a/content_delegate.js
+++ b/content_delegate.js
@@ -61,14 +61,19 @@ ContentDelegate.prototype.findPacerDocIds = function() {
       //   function goDLS(hyperlink, de_caseid, de_seqno, got_receipt,
       //     pdf_header, pdf_toggle_possible, magic_num, hdr) {
 
-      let goDLS = onclick.match(/^goDLS\('([^']*)','([^']*)','([^']*)','([^']*)','([^']*)','([^']*)','([^']*)','([^']*)'\)/);
+      let goDLS;
+      let [, hyperlink, de_caseid, de_seqno, got_receipt,
+           pdf_header, pdf_toggle_possible, magic_num, hdr ]
+          = goDLS
+          = onclick.match(/^goDLS\('([^']*)','([^']*)','([^']*)','([^']*)','([^']*)','([^']*)','([^']*)','([^']*)'\)/);
 
-      if (goDLS[2]) {
-	docsToCases[pacer_doc_id] = goDLS[2];
-	// console.info('RECAP debug: Y doc ' + pacer_doc_id + ' to ' + goDLS[2]);
+      if (de_caseid) {
+        docsToCases[pacer_doc_id] = de_caseid;
+        // console.info('RECAP debug: Y doc ' + pacer_doc_id + ' to ' + de_caseid);
       } else if (page_pacer_case_id) {
-	docsToCases[pacer_doc_id] = page_pacer_case_id;
-	// console.info('RECAP debug: X doc ' + pacer_doc_id + ' to ' + page_pacer_case_id);
+        docsToCases[pacer_doc_id] = page_pacer_case_id;
+        // console.info('RECAP debug: X doc ' + pacer_doc_id + ' to '
+        //              + page_pacer_case_id);
       }
     }
   }
--- a/content_delegate.js
+++ b/content_delegate.js
@@ -78,7 +78,7 @@ ContentDelegate.prototype.findPacerDocIds = function() {
     }
   }
 
-  chrome.storage.local.set(docsToCases, function(){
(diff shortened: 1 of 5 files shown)


pacer.js
```diff
@@ -113,11 +113,17 @@ let PACER = {
     for (let url of urls) {
       let hostname = getHostname(url);
       // JS is trash. It lacks a way of getting the TLD, so we use endsWith.
+      //
+      // xxx aren't we only invoked when we match the "matches" and
+      // "include_globs" in manifest.json, which limits us to
+      // *://*.uscourts.gov/* ?
```
mlissner: The comment on line 112  alludes to this, but this function also parses referers, which can have other domains in them. The other way to do this would be to make the match more specific than "find me some numbers after a question mark at the end of a string." But I think it's OK as is?

content_delegate.js
```diff
@@ -25,24 +25,59 @@ ContentDelegate.prototype.findPacerDocIds = function() {
...
+      // </a>
+      // The parameters are defined in the unminified js
+      //   https://ecf.flnd.uscourts.gov/lib/dls_url.js
+      // as:
+      //   function goDLS(hyperlink, de_caseid, de_seqno, got_receipt,
+      //     pdf_header, pdf_toggle_possible, magic_num, hdr) {
+
+      let goDLS = onclick.match(/^goDLS\('([^']*)','([^']*)','([^']*)','([^']*)','([^']*)','([^']*)','([^']*)','([^']*)'\)/);
+
+      if (goDLS[2]) {
```
mlissner: I think a comment explaining this bit here would be useful.

pacer.js
```diff
@@ -113,11 +113,17 @@ let PACER = {
...
       if (hostname.endsWith('uscourts.gov')) {
-        let match = url.match(/\?(\d+)$/);
-        if (match) {
+        let match;
+        if (match = url.match(/\?(\d+)$/)) {
```
mlissner: I just noticed, but I think this still needs a tweak to remove the `$` on the end (like you did above). Otherwise, a URL like this won't work:

https://ecf.almd.uscourts.gov/cgi-bin/DktRpt.pl?38115&

I just stumbled on this because of this ticket. Looks like we can fix it too while we're in here: https://github.com/freelawproject/recap/issues/168

content_delegate.js
johnhawkinson: Clarified via ["array destructuring"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Unpacking_values_from_a_regular_expression_match)

content_delegate.js
```diff
@@ -25,24 +25,59 @@ ContentDelegate.prototype.findPacerDocIds = function() {
...
+	// console.info('RECAP debug: X doc ' + pacer_doc_id + ' to ' + page_pacer_case_id);
+      }
     }
   }
-  let pacer_case_id = this.pacer_case_id ||
-    this.recap.getPacerCaseIdFromPacerDocId(this.pacer_doc_id, function(){});
-  if (pacer_case_id){
-    // If we have the case ID or have it cached for the URL, stash a mapping of
-    // the doc ids on the page to the case id we were able to find.
-    this.recap.storePacerDocIds(this.pacer_doc_ids, pacer_case_id, function(){
```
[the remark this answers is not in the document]
johnhawkinson: Yes. IMNSHO it's not a great idea that we throw our random document/caseid mapping as toplevel key/value pairs in the storage object. What if there are other mappings that conflict in the future? Or, unlikely as it seems, a doc1 number called "options"?

We should reserve the right to change this API in the future and thus maintain the abstraction.
As you know, this PR was put together from stuff that I did not have ready to commit, and this was one of the items. Sorry I missed it's half-bakedness on initial review.

manifest.json
```diff
@@ -73,6 +73,5 @@
   },
   "web_accessible_resources": [
     "assets/images/*.png",
-    "assets/js/*.js"
```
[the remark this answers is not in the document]
johnhawkinson: Ironically we both missed that I introduced a syntax error [trailing comma].
Fixed.

pacer.js
johnhawkinson: Right. clarified.

pacer.js
johnhawkinson: Sure. I was trying to do the minimal change necessary to support the feature I was working on, but we can absolutely relax this, done.

content_delegate.js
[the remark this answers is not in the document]
mlissner: Agree. It'd be good to not have this at the root.

PR #35, opened 2017-12-12
Miscellaneous cleanup from Dec. 11, 2017 commits

johnhawkinson:
Keep regexps on a single line.
Ensure terminal comma on the final line of multiline arrays so future patches are clean.
Ensure statements end in semicolons.
Correct log message from a99f6562f1f646f6508202870035027c70200ede.
Clarify comment regarding Appellate use of caseId.
Consistently use goDLS over goDls (either way we must be consistent. There's a good argument for this one though, I think).
Eliminate unnecessary temp variable.
String.prototype.match() can fail, and when it does, it will cause a TypeError from array destructuring if we don't check. Correct that.
Be smart when destructuring to an object -- don't use a pile of temp vars and repeat them, just destructure into the object directly.
debuglog the source of caseids per discussion at d8754eb#commitcomment-26188087
Match the `caseid` parameter case-insensitively per same

### DIFF
--- a/content_delegate.js
+++ b/content_delegate.js
@@ -49,10 +49,9 @@ ContentDelegate.prototype.findAndStorePacerDocIds = function() {
 
       let onclick = link.getAttribute('onclick');
       let goDLS = PACER.parseGoDLSFunction(onclick);
-      let de_caseid = goDLS.de_caseid;
-      if (de_caseid) {
-        docsToCases[pacer_doc_id] = de_caseid;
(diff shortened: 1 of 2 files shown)


mlissner:
Thanks. Great stuff. Especially appreciate the destructuring bit. I stared at that a few minutes, googled it briefly, then figured I could live with a little redundancy until my JS got good enough to figure it out intuitively. Solution makes sense in hindsight.

PR #41, opened 2018-01-19
Fixes claim doc1 URLs 

mlissner:
Prior to this fix, we'd try to send bankruptcy claim documents to the RECAP Archive. Normally we want all the documents in PACER, but these documents are an example of documents that we *don't* want since they're not even part of the docket. 

I wouldn't normally worry too much about an esoteric type of document on an esoteric type of the legal world (bankruptcy stuff), but this particular bug made it so that the document itself couldn't be downloaded at all. Breaking PACER is not OK, so this fixes that.

### DIFF
--- a/content_delegate.js
+++ b/content_delegate.js
@@ -228,10 +228,6 @@ ContentDelegate.prototype.onDocumentViewSubmit = function (event) {
   // Save a copy of the page source, altered so that the "View Document"
   // button goes forward in the history instead of resubmitting the form.
   let form = document.getElementById(event.data.id);
-  let originalSubmit = form.getAttribute('onsubmit');
-  form.setAttribute('onsubmit', 'history.forward(); return false;');
-  let previousPageHtml = document.documentElement.innerHTML;
-  form.setAttribute('onsubmit', originalSubmit);
 
   // Grab the document number, attachment number, and docket number
   let document_number, attachment_number, docket_number;
@@ -240,12 +236,22 @@ ContentDelegate.prototype.onDocumentViewSubmit = function (event) {
     let image_string = $('td:contains(Image)').text();
     let regex = /(\d+)-(\d+)/;
     let matches = regex.exec(image_string);
+    if (!matches) {
+      // This happens on bankruptcy claim pages, which are a form of document in
+      // CM/ECF that we don't want. In this case, just submit the original form.
+      form.submit();
+    }
     document_number = matches[1];
     attachment_number = matches[2];
     docket_number = $.trim($('tr:contains(Case Number) td:nth(1)').text());
   } else { // Appellate
     debug(4,"Appellate parsing not yet implemented");
   }
+
+  let originalSubmit = form.getAttribute('onsubmit');
+  form.setAttribute('onsubmit', 'history.forward(); return false;');
+  let previousPageHtml = document.documentElement.innerHTML;
+  form.setAttribute('onsubmit', originalSubmit);
 
   // Now do the form request to get to the view page.  Some PACER sites will
   // return an HTML page containing an <iframe> that loads the PDF document;
--- a/content_delegate.js
+++ b/content_delegate.js
@@ -227,7 +227,11 @@ ContentDelegate.prototype.handleSingleDocumentPageCheck = function() {
 ContentDelegate.prototype.onDocumentViewSubmit = function (event) {
   // Save a copy of the page source, altered so that the "View Document"
   // button goes forward in the history instead of resubmitting the form.
-  let form = document.getElementById(event.data.id);
+  let originalForm = document.forms[0];
+  let originalSubmit = originalForm.getAttribute('onsubmit');
+  originalForm.setAttribute('onsubmit', 'history.forward(); return false;');
+  let previousPageHtml = document.documentElement.innerHTML;
+  originalForm.setAttribute('onsubmit', originalSubmit);
 
   // Grab the document number, attachment number, and docket number
   let document_number, attachment_number, docket_number;
@@ -248,16 +252,12 @@ ContentDelegate.prototype.onDocumentViewSubmit = function (event) {
     debug(4,"Appellate parsing not yet implemented");
   }
 
-  let originalSubmit = form.getAttribute('onsubmit');
-  form.setAttribute('onsubmit', 'history.forward(); return false;');
-  let previousPageHtml = document.documentElement.innerHTML;
-  form.setAttribute('onsubmit', originalSubmit);
-
   // Now do the form request to get to the view page.  Some PACER sites will
   // return an HTML page containing an <iframe> that loads the PDF document;
   // others just return the PDF document.  As we don't know whether we'll get
   // HTML (text) or PDF (binary), we ask for an ArrayBuffer and convert later.
   $('body').css('cursor', 'wait');
+  let form = document.getElementById(event.data.id);
   let data = new FormData(form);
   httpRequest(form.action, data, function (type, ab, xhr) {
     console.info('RECAP: Successfully submitted RECAP "View" button form: ' +
--- a/content_delegate.js
+++ b/content_delegate.js
@@ -240,11 +240,6 @@ ContentDelegate.prototype.onDocumentViewSubmit = function (event) {
     let image_string = $('td:contains(Image)').text();
     let regex = /(\d+)-(\d+)/;
     let matches = regex.exec(image_string);
-    if (!matches) {
-      // This happens on bankruptcy claim pages, which are a form of document in
-      // CM/ECF that we don't want. In this case, just submit the original form.
-      form.submit();
-    }
     document_number = matches[1];
     attachment_number = matches[2];
     docket_number = $.trim($('tr:contains(Case Number) td:nth(1)').text());
--- a/content_delegate.js
+++ b/content_delegate.js
@@ -233,13 +233,20 @@ ContentDelegate.prototype.onDocumentViewSubmit = function (event) {
   let previousPageHtml = document.documentElement.innerHTML;
   originalForm.setAttribute('onsubmit', originalSubmit);
 
+  let form = document.getElementById(event.data.id);
+
   // Grab the document number, attachment number, and docket number
   let document_number, attachment_number, docket_number;
 
   if (!PACER.isAppellateCourt(this.court)) {
+    // This CSS selector duplicated in isSingleDocumentPage
     let image_string = $('td:contains(Image)').text();
     let regex = /(\d+)-(\d+)/;
     let matches = regex.exec(image_string);
+    if (!matches){
+      form.submit();
+      return;
+    }
     document_number = matches[1];
     attachment_number = matches[2];
     docket_number = $.trim($('tr:contains(Case Number) td:nth(1)').text());
@@ -252,7 +259,6 @@ ContentDelegate.prototype.onDocumentViewSubmit = function (event) {
   // others just return the PDF document.  As we don't know whether we'll get
   // HTML (text) or PDF (binary), we ask for an ArrayBuffer and convert later.
   $('body').css('cursor', 'wait');
-  let form = document.getElementById(event.data.id);
   let data = new FormData(form);
   httpRequest(form.action, data, function (type, ab, xhr) {
     console.info('RECAP: Successfully submitted RECAP "View" button form: ' +
--- a/content_delegate.js
+++ b/content_delegate.js
@@ -227,11 +227,11 @@ ContentDelegate.prototype.handleSingleDocumentPageCheck = function() {
 ContentDelegate.prototype.onDocumentViewSubmit = function (event) {
   // Save a copy of the page source, altered so that the "View Document"
   // button goes forward in the history instead of resubmitting the form.
-  let originalSubmit = document.forms[0].getAttribute('onsubmit');
-  document.forms[0].setAttribute('onsubmit',
-				 'history.forward(); return false;');
+  let form = document.getElementById(event.data.id);
+  let originalSubmit = form.getAttribute('onsubmit');
+  form.setAttribute('onsubmit', 'history.forward(); return false;');
   let previousPageHtml = document.documentElement.innerHTML;
-  document.forms[0].setAttribute('onsubmit', originalSubmit);
+  form.setAttribute('onsubmit', originalSubmit);
 
   // Grab the document number, attachment number, and docket number
   let document_number, attachment_number, docket_number;
@@ -252,7 +252,6 @@ ContentDelegate.prototype.onDocumentViewSubmit = function (event) {
   // others just return the PDF document.  As we don't know whether we'll get
   // HTML (text) or PDF (binary), we ask for an ArrayBuffer and convert later.
   $('body').css('cursor', 'wait');
-  let form = document.getElementById(event.data.id);
   let data = new FormData(form);
   httpRequest(form.action, data, function (type, ab, xhr) {
     console.info('RECAP: Successfully submitted RECAP "View" button form: ' +
--- a/pacer.js
+++ b/pacer.js
@@ -139,7 +139,11 @@ let PACER = {
   isSingleDocumentPage: function (url, document) {
     let inputs = document.getElementsByTagName('input');
     let lastInput = inputs.length && inputs[inputs.length - 1].value;
+    // If the receipt doesn't say "Image" we don't yet support it on the server.
+    // So far, this only appears to apply to bankruptcy claims.
+    let hasImageReceipt = !!$('td:contains(Image)').length;
     let pageCheck = (PACER.isDocumentUrl(url) &&
+                     hasImageReceipt &&
                      (lastInput === 'View Document') ||
                      (lastInput === 'Accept Charges and Retrieve'));
     debug(4," lastInput "+lastInput);
--- a/pacer.js
+++ b/pacer.js
@@ -140,7 +140,8 @@ let PACER = {
     let inputs = document.getElementsByTagName('input');
     let lastInput = inputs.length && inputs[inputs.length - 1].value;
     // If the receipt doesn't say "Image" we don't yet support it on the server.
-    // So far, this only appears to apply to bankruptcy claims.
+    // So far, this only appears to apply to bankruptcy claims. This CSS
+    // selector is duplicated in onDocumentViewSubmit.
     let hasImageReceipt = !!$('td:contains(Image)').length;
     let pageCheck = (PACER.isDocumentUrl(url) &&
                      hasImageReceipt &&
--- a/spec/ContentDelegateSpec.js
+++ b/spec/ContentDelegateSpec.js
@@ -339,6 +339,14 @@ describe('The ContentDelegate class', function() {
         input = document.createElement('input');
         input.value = 'View Document';
         form.appendChild(input);
+
+        table = document.createElement('table');
+        tr = document.createElement('tr');
+        td = document.createElement('td');
+        td.appendChild(document.createTextNode('Image'));
+        tr.appendChild(td);
+        table.appendChild(tr);
+        document.body.appendChild(table);
       });
 
       afterEach(function() {
@@ -429,6 +437,14 @@ describe('The ContentDelegate class', function() {
         input = document.createElement('input');
         input.value = 'View Document';
         form.appendChild(input);
+
+        table = document.createElement('table');
+        tr = document.createElement('tr');
+        td = document.createElement('td');
+        td.appendChild(document.createTextNode('Image'));
+        tr.appendChild(td);
+        table.appendChild(tr);
+        document.body.appendChild(table);
       });
 
       afterEach(function() {
@@ -663,7 +679,7 @@ describe('The ContentDelegate class', function() {
         expect(window_obj.location).toBe(linkUrl);
       });
     });
-    
+
     describe('when the popup option is set', function() {
       beforeEach(function() {
         window.chrome =  {
@@ -739,7 +755,7 @@ describe('The ContentDelegate class', function() {
         cd.attachRecapLinkToEligibleDocs();
         expect($('.recap-inline').length).toBe(0);
       });
-      
+
       it('attaches a single link to the one url with recap', function() {
         spyOn(cd.recap, 'getAvailabilityForDocuments').and.callFake(
           function(_, _, callback) {


content_delegate.js
```diff
@@ -227,11 +227,7 @@ ContentDelegate.prototype.handleSingleDocumentPageCheck = function() {
 ContentDelegate.prototype.onDocumentViewSubmit = function (event) {
   // Save a copy of the page source, altered so that the "View Document"
   // button goes forward in the history instead of resubmitting the form.
-  let originalSubmit = document.forms[0].getAttribute('onsubmit');
```
audiodude: So the main difference here is that we will attach to the form of the event rather than just generically to the first form on the page. Is that what you're trying to do? Seems more precise, but definitely a change in behavior potentially.

mlissner: Yeah, they should be the same thing. I'm not entirely sure just by reading the code, but the functionality still seems to be working properly.

content_delegate.js
```diff
@@ -227,11 +227,7 @@ ContentDelegate.prototype.handleSingleDocumentPageCheck = function() {
...
-  document.forms[0].setAttribute('onsubmit',
-				 'history.forward(); return false;');
-  let previousPageHtml = document.documentElement.innerHTML;
-  document.forms[0].setAttribute('onsubmit', originalSubmit);
+  let form = document.getElementById(event.data.id);
```
johnhawkinson: To feel confident that this is right, I'd want to know that `document.forms[0]` and `document.getElementById(event.data.id)` resolve to the same form. I don't have that confidence from inspecting the code. And when I look at PACER forms with the DOM inspector, etc., I don't see an `id` property. And when I set a breakpoint at the top of `onDocumentViewSubmit()`, I also don't get that.

For the Bankruptcy case, which I guess is the brokenness this is trying to solve, they are certainly not the same:

```
>> event
message { target: Window → 032026507697, isTrusted: false, data: Object, origin: "https://ecf.caeb.uscourts.gov", lastEventId: "", source: Window → 032026507697, ports: Object, currentTarget: Window → 032026507697, eventPhase: 2, bubbles: false, cancelable: false }
>> event.data
Object { id: "form1516491714124" }
>> event.data.id
"form1516491714124"
 form = document.getElementById(event.data.id);
<form action="/doc1/032126507697" method="post" id="form1516491714124">
>> form.outerHTML
"<form action="/doc1/032126507697" method="post" id="form1516491714124"><input name="caseid" value="605035" type="hidden"><input name="got_receipt" value="1" type="hidden"></form>"
>> document.forms
HTMLCollection [ <form>, <form#form1516491714124> ]
>> document.forms.length
2
>> document.forms[0]
<form method="POST" action="https://ecf.caeb.uscourts.gov/doc1/032126507697" onsubmit="goDLS('/doc1/032126507697','605035','','1','','','','','','');return(false)">
>> document.forms[0].innerHTML
"<hr><center><table width="400" border="1" bgcolor="white"><tbody><tr><th colspan="4"><font size="+1" color="DARKRED">PACER Service Center </font></th></tr><tr><th colspan="4"><font color="DARKBLUE">Transaction Receipt </font></th></tr><tr></tr><tr></tr><tr><td colspan="4" align="CENTER"><font size="-1" color="DARKBLUE">Sat Jan 20 15:41:41 2018</font></td></tr><tr><th align="LEFT"><font size="-1" color="DARKBLUE"> Pacer Login: </font></th><td align="LEFT"><font size="-1" color="DARKBLUE"> jhawkinson0:3355704:0 </font></td><th align="LEFT"><font size="-1" color="DARKBLUE"> Client Code: </font></th><td align="LEFT"><font size="-1" color="DARKBLUE">  </font></td></tr><tr><th align="LEFT"><font size="-1" color="DARKBLUE"> Description: </font></th><td align="LEFT"><font size="-1" color="DARKBLUE"> Image:72-0 </font></td><th align="LEFT"><font size="-1" color="DARKBLUE"> Case Number: </font></th><td align="LEFT"><font size="-1" color="DARKBLUE"> 17-13797                       </font></td></tr><tr><th align="LEFT"><font size="-1" color="DARKBLUE"> Billable Pages: </font></th><td align="LEFT"><font size="-1" color="DARKBLUE"> 2 </font></td><th align="LEFT"><font size="-1" color="DARKBLUE"> Cost: </font></th><td align="LEFT"><font size="-1" color="DARKBLUE"> 0.20 </font></td></tr><tr></tr><tr></tr></tbody></table></center><input value="View Document" type="submit">"
>> document.forms[1].innerHTML
"<input name="caseid" value="605035" type="hidden"><input name="got_receipt" value="1" type="hidden">"

```

But I see the same with CMECF District now, which was not broken before, and maybe is with this code?:

```
>> document.forms.length
2
>> event.data.id
"form1516491978507"
>> document.forms
HTMLCollection [ <form>, <form#form1516491978507> ]
>> document.forms[0].outerHTML
"<form method="post" action="https://ecf.mad.uscourts.gov/doc1/09518488434" onsubmit="goDLS('/doc1/09518488434','192392','666','1','','1','','');return(false);"><hr><center><table width="400" border="1" bgcolor="white"><tbody><tr><th colspan="4"><font size="+1" color="DARKRED">Pacer Service Center </font></th></tr><tr><th colspan="4"><font color="DARKBLUE">Transaction Receipt </font></th></tr><tr></tr><tr></tr><tr><td colspan="4" align="CENTER"><font size="-1" color="DARKBLUE">Sat Jan 20 18:46:00 2018</font></td></tr><tr><th align="LEFT"><font size="-1" color="DARKBLUE"> Pacer Login: </font></th><td align="LEFT"><font size="-1" color="DARKBLUE"> jhawkinson0:3355704:0 </font></td><th align="LEFT"><font size="-1" color="DARKBLUE"> Client Code: </font></th><td align="LEFT"><font size="-1" color="DARKBLUE">  </font></td></tr><tr><th align="LEFT"><font size="-1" color="DARKBLUE"> Description: </font></th><td align="LEFT"><font size="-1" color="DARKBLUE"> Image83-0 </font></td><th align="LEFT"><font size="-1" color="DARKBLUE"> Case Number: </font></th><td align="LEFT"><font size="-1" color="DARKBLUE"> 1:17-cv-11842-PBS</font></td></tr><tr><th align="LEFT"><font size="-1" color="DARKBLUE"> Billable Pages: </font></th><td align="LEFT"><font size="-1" color="DARKBLUE"> 2 </font></td><th align="LEFT"><font size="-1" color="DARKBLUE"> Cost: </font></th><td align="LEFT"><font size="-1" color="DARKBLUE"> 0.20 </font></td></tr><tr></tr><tr></tr></tbody></table></center><input value="View Document" type="submit"></form>"
>> document.forms[1].outerHTML
"<form action="/doc1/09518488434" enctype="multipart/form-data" method="post" id="form1516491978507"><input name="caseid" value="192392" type="hidden"><input name="de_seq_num" value="666" type="hidden"><input name="got_receipt" value="1" type="hidden"><input name="pdf_toggle_possible" value="1" type="hidden"></form>"
```

content_delegate.js
johnhawkinson: I think the functionality that the `onsubmit` / `history.forward()` / `previousPageHtml` is dealing with is the ability to hit Back and have The Right Thing Happen. Is that what you tested, Mike? Or did you just test that downloads worked?

content_delegate.js
```diff
@@ -240,19 +236,28 @@ ContentDelegate.prototype.onDocumentViewSubmit = function (event) {
     let image_string = $('td:contains(Image)').text();
```
johnhawkinson: Can we just make this code:

```js
let document_string = $('th:contains(Description:)').next().text()
```

so that it handles both the DC case
```html
<tr>
<th align="LEFT"><font size="-1" color="DARKBLUE"> Description: </font></th>
<td align="LEFT"><font size="-1" color="DARKBLUE"> Image83-0 </font>
```

and also the BK case:
```html
<tr>
<th align="LEFT"><font size="-1" color="DARKBLUE"> Description: </font></th>
<td align="LEFT"><font size="-1" color="DARKBLUE"> Claim 43976411-0 </font></td>
```

content_delegate.js
mlissner: I tested both, yeah. If you download a claim and then hit back, it works.

content_delegate.js
mlissner: Well, when it says Claim, ideally we want that *not* to get parsed anymore. That's a tip off that we're in a place we don't want to be doing uploads.

content_delegate.js
```diff
@@ -240,19 +236,28 @@ ContentDelegate.prototype.onDocumentViewSubmit = function (event) {
...
     let regex = /(\d+)-(\d+)/;
     let matches = regex.exec(image_string);
+    if (!matches) {
```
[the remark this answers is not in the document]
mlissner: Yeah, this occurred to me too earlier today. Might be worth adding a check there instead. I'll look into that.

content_delegate.js
```diff
@@ -240,13 +236,23 @@ ContentDelegate.prototype.onDocumentViewSubmit = function (event) {
...
+      // This happens on bankruptcy claim pages, which are a form of document in
```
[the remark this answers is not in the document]
mlissner: I'll tweak (if this code remains in the final version.)

content_delegate.js
```diff
@@ -240,13 +236,23 @@ ContentDelegate.prototype.onDocumentViewSubmit = function (event) {
...
+      // CM/ECF that we don't want. In this case, just submit the original form.
+      form.submit();
```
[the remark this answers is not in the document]
mlissner: Huh. Yeah. Interesting that it worked regardless.

content_delegate.js
```diff
@@ -240,13 +236,23 @@ ContentDelegate.prototype.onDocumentViewSubmit = function (event) {
...
+    }
     document_number = matches[1];
     attachment_number = matches[2];
     docket_number = $.trim($('tr:contains(Case Number) td:nth(1)').text());
   } else { // Appellate
     debug(4,"Appellate parsing not yet implemented");
   }
 
+  let originalSubmit = form.getAttribute('onsubmit');
```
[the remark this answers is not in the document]
mlissner: Agree. Will fix.

content_delegate.js
johnhawkinson: Wellllll... **_ideally_**? It seems…quite a bit removed from _ideally_.

1. Ideally, RECAP should mirror absolutely all of PACER, but we're not there yet. We'll probably never handle deadline dates or case file locations, etc.
2. Practically, RECAP should handle all objects of a particular kind in PACER, not pick and choose. For instance, it should handle all PDFs. As far as I/we know, all PACER PDFs have a DLS and are "Documents" with receipts that belong to a case with a docket number and a `pacer_case_id`. This is part of why I get so incensed about freelawproject/recap#61  and the idea that the server accepts PDFs but withholds them from users pending metadata. That's like…cruel and miserly. So what if there's no metadata? That shouldn't justify my having to pay $3 for a document that someone else has crowdsourced and contributed!
3. However, some of them are Claims, not Images, and belong to a "Claims Register" not a "Docket Report." That requires both server-side and client-side support, which is not on the roadmap.

But it seems to me we are much better off sending them to the server now than holding them back deliberately. It gets us closer to where we want, _ideally_, to be.

Now, maybe something goes wrong if we do that, further down? The obvious is that **(a)** `document_number` and `attachment_number` may not be semantically sane, and **(b)** we are not passing the PDF type (Image or Claim) to `showPdfPage()` and `recap.uploadDocument()`.

I'm not sure (a) is true. So what if we upload `{document_number: "43976411", attachment_number: "0"}` and not `{claim_number: "70-0"}`? Yes, the document number isn't nice and sequential like in DC docket reports, but we already have this in CMECF Appellate and it's fine there (for the IA imported data) so it should be fine here too. And will the `attachment_number` get lost because it's not truthy and we do we care? I don't see why we must care (although perhaps we should).

And I'm not sure (b) is required. Yeah, we should probably adjust the regexp to capture the doctype and pass it to the server, even if the server currently throws it away. And yeah, we should probably fix the server to save it because maybe it'll come in handy later when somebody wants to add support for the Claims Register. 

**(c)** We could also grab the claim number from the Receipt page url (e.g. 
`/cgi-bin/show_doc.pl?caseid=605035&claim_id=43976411&claim_num=70-1&magic_num=MAGIC`) and pass that on. Not required, but probably smart to do, but not on the roadmap and not necessary for PDF crowdsourcing.

But I'm not seeing why we should take steps to make that any harder than it has to be, or to encourage preclusion like this.

johnhawkinson:
Just to be clear, this PR addresses freelawproject/recap#234
(I kept looking for the link and not finding it, so here it is.)

content_delegate.js
johnhawkinson: Whoops, correction, although it may not matter. The above BK example was from an image document in the docket report, but the proper test case would be from a claim document in the claims register. It seems the same with respect to which form is which, though:

```
>> document.forms
HTMLCollection [ <form>, <form#form1516523730586> ]
>> document.forms[0].outerHTML
"<form method="POST" action="https://ecf.caeb.uscourts.gov/doc1/032126913922" onsubmit="goDLS('/doc1/032126913922','605035','','1','2','','','43976411','70-1','');return(false)"><hr><center><table width="400" border="1" bgcolor="white"><tbody><tr><th colspan="4"><font size="+1" color="DARKRED">PACER Service Center </font></th></tr><tr><th colspan="4"><font color="DARKBLUE">Transaction Receipt </font></th></tr><tr></tr><tr></tr><tr><td colspan="4" align="CENTER"><font size="-1" color="DARKBLUE">Sun Jan 21 00:34:29 2018</font></td></tr><tr><th align="LEFT"><font size="-1" color="DARKBLUE"> Pacer Login: </font></th><td align="LEFT"><font size="-1" color="DARKBLUE"> jhawkinson0:3355704:0 </font></td><th align="LEFT"><font size="-1" color="DARKBLUE"> Client Code: </font></th><td align="LEFT"><font size="-1" color="DARKBLUE">  </font></td></tr><tr><th align="LEFT"><font size="-1" color="DARKBLUE"> Description: </font></th><td align="LEFT"><font size="-1" color="DARKBLUE"> Claim 43976411-0 </font></td><th align="LEFT"><font size="-1" color="DARKBLUE"> Case Number: </font></th><td align="LEFT"><font size="-1" color="DARKBLUE"> 17-13797                       </font></td></tr><tr><th align="LEFT"><font size="-1" color="DARKBLUE"> Billable Pages: </font></th><td align="LEFT"><font size="-1" color="DARKBLUE"> 30 </font></td><th align="LEFT"><font size="-1" color="DARKBLUE"> Cost: </font></th><td align="LEFT"><font size="-1" color="DARKBLUE"> 3.00 </font></td></tr><tr></tr><tr></tr></tbody></table></center><input value="View Document" type="submit"></form>"
>> document.forms[1].outerHTML
"<form action="/doc1/032126913922" method="post" id="form1516523730586"><input name="caseid" value="605035" type="hidden"><input name="got_receipt" value="1" type="hidden"><input name="pdf_header" value="2" type="hidden"><input name="claim_id" value="43976411" type="hidden"><input name="claim_num" value="70-1" type="hidden"></form>"
```

Maybe it's wise to make that more readable:
```html
<form method="POST" action="https://ecf.caeb.uscourts.gov/doc1/032126913922" onsubmit="goDLS('/doc1/032126913922','605035','','1','2','','','43976411','70-1','');return(false)">
<hr><center>
<table width="400" border="1" bgcolor="white"><tbody>
<tr><th colspan="4"><font size="+1" color="DARKRED">PACER Service Center </font></th></tr>
<tr><th colspan="4"><font color="DARKBLUE">Transaction Receipt </font></th></tr>
<tr></tr>
<tr></tr>
<tr><td colspan="4" align="CENTER"><font size="-1" color="DARKBLUE">Sun Jan 21 00:34:29 2018</font></td></tr>
<tr><th align="LEFT"><font size="-1" color="DARKBLUE"> Pacer Login: </font></th><td align="LEFT"><font size="-1" color="DARKBLUE"> jhawkinson0:3355704:0 </font></td>
<th align="LEFT"><font size="-1" color="DARKBLUE"> Client Code: </font></th><td align="LEFT"><font size="-1" color="DARKBLUE">  </font></td></tr>
<tr><th align="LEFT"><font size="-1" color="DARKBLUE"> Description: </font></th><td align="LEFT"><font size="-1" color="DARKBLUE"> Claim 43976411-0 </font>
</td><th align="LEFT"><font size="-1" color="DARKBLUE"> Case Number: </font></th><td align="LEFT"><font size="-1" color="DARKBLUE"> 17-13797                       </font></td></tr>
<tr><th align="LEFT"><font size="-1" color="DARKBLUE"> Billable Pages: </font></th><td align="LEFT"><font size="-1" color="DARKBLUE"> 30 </font></td>
<th align="LEFT"><font size="-1" color="DARKBLUE"> Cost: </font></th><td align="LEFT"><font size="-1" color="DARKBLUE"> 3.00 </font></td></tr>
<tr></tr>
<tr></tr>
</tbody></table></center>
<input value="View Document" type="submit">
</form>
```
and
```html
<form action="/doc1/032126913922" method="post" id="form1516523730586">
<input name="caseid" value="605035" type="hidden">
<input name="got_receipt" value="1" type="hidden">
<input name="pdf_header" value="2" type="hidden">
<input name="claim_id" value="43976411" type="hidden">
<input name="claim_num" value="70-1" type="hidden">
</form>
```

> To feel confident that this is right, I'd want to know that document.forms[0] and document.getElementById(event.data.id) resolve to the same form. I don't have that confidence from inspecting the code. And when I look at PACER forms with the DOM inspector, etc., I don't see an id property. And when I set a breakpoint at the top of onDocumentViewSubmit(), I also don't get that.

So, knowing they resolve to the same form may be too high a standard. It should be sufficient if both forms' `onsubmit` actions get triggered in nearly the same browser page history context.  I'm not sure that's a true statement either, but it seems an easier reach?

content_delegate.js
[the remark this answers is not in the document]
johnhawkinson: Well, if you submit a form twice, it'll "work." (Maybe you get charged twice? Although I've always wondered if PACER is forgiving if you download the same PDF 2x in a row, e.g., because network problems interrupted the first attempt).

content_delegate.js
johnhawkinson: Hrmmm....I wonder if (a) it is browser-specific (b) if this code is even necessary at all?

content_delegate.js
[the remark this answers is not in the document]
mlissner: Well, what's weird is the next line is this: 

    document_number = matches[1];

Which would have crashed things if it was reached, since that's where we began this bug.

mlissner:
OK, I just pushed a new approach to this problem. It:

 - Short circuits the code so that claims pages are not identified as document pages anymore.
 - Restores previous functionality WRT `forms[0]` vs. `event.data.id`.

content_delegate.js
mlissner: I tested this and it worked fine, but I reviewed it again today and realized there was indeed a change in functionality that this introduced. I backed out the changes. Even if it works, I don't mean to be tinkering here right now.

pacer.js
```diff
@@ -139,7 +139,11 @@ let PACER = {
   isSingleDocumentPage: function (url, document) {
     let inputs = document.getElementsByTagName('input');
     let lastInput = inputs.length && inputs[inputs.length - 1].value;
+    // If the receipt doesn't say "Image" we don't yet support it on the server.
+    // So far, this only appears to apply to bankruptcy claims.
+    let hasImageReceipt = !!$('td:contains(Image)').length;
```
johnhawkinson: This is OK, but I am troubled by by the duplication of `$('td:contains(Image)').length` here with `$('td:contains(Image)').text()` in `content_delegate.js`. But the only obvious way I see to address that would be to store the value of `$('td:contains(Image)').text()` in some page state, like `this.imageString`, but that works poorly because `this` in `pacer.js` is not the same as `this` in `content_delegate.js` and there's not a great way for `PACER` functions to adjust `content_delegate` state, as far as I know. I guess `isDocumentSinglePage` could take `(url, document, that)` instead of `(url, document)` but this all gets messy and I know you're trying to cut this quickly.

So perhaps a comment here (and maybe in the other one) to note the duplication and that they need to stay in sync.

spec/ContentDelegateSpec.js
```diff
@@ -663,7 +679,7 @@ describe('The ContentDelegate class', function() {
         expect(window_obj.location).toBe(linkUrl);
       });
     });
-
```
johnhawkinson: Whitespace changes in their own commit, please.

spec/ContentDelegateSpec.js
```diff
@@ -739,7 +755,7 @@ describe('The ContentDelegate class', function() {
         cd.attachRecapLinkToEligibleDocs();
         expect($('.recap-inline').length).toBe(0);
       });
-
```
johnhawkinson: Ibid.

content_delegate.js
```diff
@@ -240,11 +240,6 @@ ContentDelegate.prototype.onDocumentViewSubmit = function (event) {
     let image_string = $('td:contains(Image)').text();
     let regex = /(\d+)-(\d+)/;
     let matches = regex.exec(image_string);
-    if (!matches) {
```
johnhawkinson: Removing this safety check is a mistake.
We should always have a plan if the regexp fails, e.g. the case number lacks a hyphen, the receipt structure is different, or there's a letter prefix in the docket number `10A-22`, or who knows.

I'd say, then, that the function should return after calling `form.submit()` rather than continuing to execute. And we know this code path is unlikely to be well-utilized, but it is a safety bound and that is a wise thing.

mlissner:
I went ahead and merged this one, thank you for all the feedback. I didn't change the variable names from `originalForm` because that was following in the convention that was already there in the `originalSubmit` variable, but I think I made the rest of the changes.

johnhawkinson:
> I went ahead and merged this one, thank you for all the feedback. I didn't change the variable names from `originalForm` because that was following in the convention that was already there in the `originalSubmit` variable, but I think I made the rest of the changes.

Just for the record, that misinterprets. We had ("originally"):

```
  let originalSubmit = document.forms[0].getAttribute('onsubmit');
  document.forms[0].setAttribute('onsubmit',
                                 'history.forward(); return false;');
  let previousPageHtml = document.documentElement.innerHTML;
  document.forms[0].setAttribute('onsubmit', originalSubmit);
```

and you moved to:

```
  let originalForm = document.forms[0];
  let originalSubmit = originalForm.getAttribute('onsubmit');
  originalForm.setAttribute('onsubmit', 'history.forward(); return false;');
  let previousPageHtml = document.documentElement.innerHTML;
  originalForm.setAttribute('onsubmit', originalSubmit);
```

`originalSubmit`  is the original value of the `onsubmit` attribute. We get it, access it by value (not by reference), and we never change it. It as the original and it always is the original.

By contrast, you make `originalForm` a reference (not a copy by value) to the form, which is already a problem. It won't always be the original because it's a reference (although if the original form never changed, then maybe it would be ok). Then call `originalForm.setAttribute()` which modifies the value of the original form, so it is no longer the original form. Yet it is called the original form. It also modifies the value of the actual form, of course, because this is by reference. And then, after making a convenience copy, you call `.setAttribute()` one more time, reverting the form to, ahem, it's _original_ value.

This code was confusing before. Now the naming is just wrong.
It makes no semantic sense.

So it is _confusing_ **and** _wrong_.

PR #44, opened 2018-03-16
Improves docs

cdale77:
Makes some small improvements to the documentation. 

https://github.com/freelawproject/recap/issues/240

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Tips
 ----
 1. You can load an "unpacked extension" from [chrome://extensions/][c] if you enable developer options in Chrome. The same [can be accomplished in Firefox][tempff] by going to [about:debugging][abtdbg], clicking "Load Temporary Add-on" and then selecting any file.
 
-1. While it's true that every court has their own customized version of PACER, there is [a PACER training site that does not charge fees][trainwreck]. You can use this if you wish to work on the system without accruing charges. If you are accruing charges while working on this extension, Free Law Project may be able to help. Let us know.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -3,40 +3,54 @@
...
+   enable developer options in Chrome. The same [can be accomplished in
+   Firefox][tempff] by going to [about:debugging][abtdbg], clicking "Load
+   Temporary Add-on" and then selecting any file.
+
+1. While it's true that every court has their own customized version of PACER,
+   there is [a PACER training site that does not charge fees][trainwreck]. You
+   can use this if you wish to work on the system without accruing charges. Note
+   that the extension is disabled on the PACER training site to avoid training
+   documents being sent in. To enable the extension you'll need to edit the
+   blacklist. If you are accruing charges while working on this extension, Free
```
cdale77: I added: `Note that the extension is disabled on the PACER training site to avoid training documents being sent in. To enable the extension you'll need to edit the blacklist.`

mlissner:
So, I think our travis problems are resolved if you pull the latest from master, but I wanted to actually verify this before we pulled it in. Looks like I was wrong. This code:

https://github.com/freelawproject/recap-chrome/blob/d0ceaddc2f56ea9544a1f90128f848115a47339a/toolbar_button.js#L54-L61

Calls:

https://github.com/freelawproject/recap-chrome/blob/bed0e7e4b243e136fc103f01cb4d683208f4e1d7/pacer.js#L41-L45

Which seems to indicate that it'd match the training site and should work there. I know this isn't what you came here for, but do you mind digging in a bit and seeing what's going on with this? I imagine if you can get this going in a debugger you'd be well positioned to do something more interesting (or to fix this).

mlissner:
I'm just going to go ahead and close this one. I think the documentation change is useful, but there's now a bunch of other random commits going on here related to tests being weirdly broken the one day you tried to commit. If you've got the git-fu to try to unwind this and resubmit it, the fix would still be welcome, but it's just a docs thing, so I'm OK letting it go. If you are!

PR #45, opened 2018-03-17
Fix tests by mocking out the saveAs method

audiodude:
According to https://github.com/karma-runner/karma/issues/2652#issuecomment-361682911

someone else had their headless tests break around the same time as us, and `saveAs` was the culprit. Sure enough, mocking this method out fixes them.


mlissner:
Thanks Travis. I think I read that ticket up to that message and then stopped one message before it. Oye. Thanks for jumping in here and helping.

audiodude:
My pleasure! Sorry I'm not available to help more on the domain specific stuff ("Courtlistener doesn't get party affinities for appellate cases" or whatnot), but really I actually know nothing about the court system and most of it flies over my head.

PR #46, opened 2018-03-20
RecapSpec: Relax tests to allow debugging

johnhawkinson:
If DEBUG=true is set in recap.js, test should not fail.

Previously, tests of recap endpoints expected uploadDocket() &c. to ~return a result~ **send a request** with debug false.
So if debug is true, tests fail. Boo.

Instead, don't require exact equality in the actualData.
Merely require it to be a superset of the `expected` data, with
jasmine.objectContaining().

@audiodude, can you please review this for sanity?

### DIFF
--- a/spec/RecapSpec.js
+++ b/spec/RecapSpec.js
@@ -157,13 +157,12 @@ describe('The Recap export module', function() {
       expected.append('upload_type', 1);
       expected.append('court', court);
       expected.append('pacer_case_id', pacer_case_id);
-      expected.append('debug', false);
       expected.append('filepath_local', new Blob(
         [html], {type: type}), filename);
 
       recap.uploadDocket(court, pacer_case_id, html, 'DOCKET', function() {});
       var actualData = jasmine.Ajax.requests.mostRecent().data();
-      expect(actualData).toEqual(expected);
+      expect(actualData).toEqual(jasmine.objectContaining(expected));
     });
   });
 
@@ -191,13 +190,12 @@ describe('The Recap export module', function() {
       expected.append('court', court);
       expected.append('pacer_case_id', pacer_case_id);
       expected.append('upload_type', 2);
-      expected.append('debug', false);
       expected.append('filepath_local', new Blob(
         [html], {type: type}), filename);
 
       recap.uploadAttachmentMenu(court, pacer_case_id, html, function() {});
       var actualData = jasmine.Ajax.requests.mostRecent().data();
-      expect(actualData).toEqual(expected);
+      expect(actualData).toEqual(jasmine.objectContaining(expected));
     });
   });
 
@@ -234,7 +232,6 @@ describe('The Recap export module', function() {
       expected.append('document_number', docnum);
       expected.append('attachment_number', attachnum);
       expected.append('upload_type', 3);
-      expected.append('debug', false);
       expected.append('filepath_local', new Blob(
         [html], {type: type}), filename);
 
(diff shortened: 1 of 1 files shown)


johnhawkinson:
To be more clear on the repro case, edit line 4 of `recap.js` to set `DEBUG` to `true` and run tests. These 2 fail.

audiodude:
So from what I can tell, you're using objectContaining to get around having to know the value of debug (true vs false).

I would say the correct way to do this is actually to write two tests, and mock the value of DEBUG in each.

mlissner:
I'm going to stay out of this one unless somebody grabs me on Slack and tells me to look at it again. Travis, I think you've got Merge rights on this repo, right?

johnhawkinson:
I think…I do not understand your use of the word "mocked," @audiodude.

> So from what I can tell, you're using objectContaining to get around having to know the value of debug (true vs false).

Yes, that is correct. It's not part of the interface.

> I would say the correct way to do this is actually to write two tests, and mock the value of DEBUG in each.

One test with DEBUG forced to true and one test with DEBUG forced to false, i.e. monkeypatching the `Recap()` object?

Why is it worth this?

audiodude:
> One test with DEBUG forced to true and one test with DEBUG forced to false, i.e. monkeypatching the Recap() object?
> 
> Why is it worth this?

The Recap code makes an AJAX request with debug set to either true or false. Presumably, this parameter has some meaning. You should not just ignore its value, you should check that it equals what you expect in each case. If the Recap code started suddenly sending debug=true for non-debug cases, or vice-versa, your test as modified in this PR would not catch it.

Currently the test is wrong, I agree. It assumes that the value of DEBUG is always true. So that half of the tests is already written (though it would be better to explicitly set Recap.DEBUG=true in those cases).

But the answer isn't to say 'The value of DEBUG doesn't matter'. If it's passing it in the AJAX request, clearly it matters. The alternative is to remove it from the AJAX request and then you won't have to test it at all.

johnhawkinson:
(Prefatory note: I incorrectly described this as testing the result of the request, but actually it's just testing the composition of the Ajax request itself.)

I think I still don't quite understand, sorry:
> The Recap code makes an AJAX request with debug set to either true or false. Presumably, this parameter has some meaning. You should not just ignore its value, you should check that it equals what you expect in each case. If the Recap code started suddenly sending debug=true for non-debug cases, or vice-versa, your test as modified in this PR would not catch it.

I think I don't get it. I mean, there are some internal machinations that ought to be in a test and there are some that aren't. For instance, if the requesting framework generated a pseudorandom nonce, or included a cryptographic key, or queried the server for a nonce and replied with that nonce, we would not expect the test to check the values of any of those things when checking the form a request that API built.

The DEBUG setting is not an API parameter. From the perspective of the callers to the RECAP code being tested here, it in fact does not have meaning. (Of course it turns out it has meaning within the code, and also on the server.) 

You're right, if the code started flipping the value of debug, we'd never know. But we also don't know what the proper value of debug is.

So:
> Presumably, this parameter has some meaning

I guess I'm not sure that's a true statement (but I'm also not sure that it isn't).

Maybe I'm thinking of these tests in the wrong way, since they're…I guess I don't know exactly. They're sort of unit tests, but not quite, because they look inside methods? And they're definitely not integration tests or functional tests. So...maybe if I better understood our testing philosophy this would be more clear to me.

I'm also still not sure on your proposed fix:
> I think…I do not understand your use of the word "mocked," ... monkeypatching the Recap() object?

Is that a Yes?

audiodude:
Based on the discussion offline, I think this is good to go. I will add a test function that tests all three methods for debug=false in a follow up.

Thanks.

johnhawkinson:
Loose ends followed up in #47.

PR #47, opened 2018-03-20
 Add tests that all Recap methods send AJAX requests with debug=false

audiodude:
Follow up from #46. The idea is that if a developer were to accidentally leave DEBUG = true in their recap.js file, the tests would break. But in general, when changing that parameter, only this test breaks.

### DIFF
--- a/spec/RecapSpec.js
+++ b/spec/RecapSpec.js
@@ -133,6 +133,42 @@ describe('The Recap export module', function() {
     });
   });
 
+  describe('uploadDocket, uploadAttachmentMenu, uploadDocument', function() {
+    var existingFormData;
+    beforeEach(function() {
+      setupChromeSpy();
+      existingFormData = window.FormData;
+      window.FormData = FormDataFake;
+    });
+    
+    afterEach(function() {
+      window.FormData = existingFormData;
+      removeChromeSpy();
+    });
+
+    var bytes = new Uint8Array([100, 100, 200, 200, 300]);
+
+    it('all send debug=false in AJAX request', function() {
+      expected = new FormDataFake();
+      expected.append('debug', false);
+
+      recap.uploadDocket(court, pacer_case_id, html, 'DOCKET', function() {});
+      var actualData = jasmine.Ajax.requests.mostRecent().data();
+      expect(actualData).toEqual(jasmine.objectContaining(expected));
+
+      recap.uploadAttachmentMenu(court, pacer_case_id, html, function() {});
+      var actualData = jasmine.Ajax.requests.mostRecent().data();
+      expect(actualData).toEqual(jasmine.objectContaining(expected));
+
+      recap.uploadDocument(
+        court, pacer_case_id, pacer_doc_id, docnum, attachnum, bytes,
+        function() {});
+      var actualData = jasmine.Ajax.requests.mostRecent().data();
+      expect(actualData).toEqual(jasmine.objectContaining(expected));
+    });
+  });
+
+
   describe('uploadDocket', function() {
     var existingFormData;
     beforeEach(function() {
--- a/spec/RecapSpec.js
+++ b/spec/RecapSpec.js
@@ -140,7 +140,7 @@ describe('The Recap export module', function() {
       existingFormData = window.FormData;
       window.FormData = FormDataFake;
     });
-    
+
     afterEach(function() {
       window.FormData = existingFormData;
       removeChromeSpy();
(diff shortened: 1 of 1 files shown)


spec/RecapSpec.js
```diff
@@ -133,6 +133,42 @@ describe('The Recap export module', function() {
...
   });
 
+  describe('uploadDocket, uploadAttachmentMenu, uploadDocument', function() {
+    var existingFormData;
+    beforeEach(function() {
+      setupChromeSpy();
+      existingFormData = window.FormData;
+      window.FormData = FormDataFake;
+    });
+
```
johnhawkinson: This whitespace at the end of line 143 should be struck.

spec/RecapSpec.js
```diff
@@ -133,6 +133,42 @@ describe('The Recap export module', function() {
...
+    });
+    
+    afterEach(function() {
+      window.FormData = existingFormData;
+      removeChromeSpy();
+    });
+
+    var bytes = new Uint8Array([100, 100, 200, 200, 300]);
+
+    it('all send debug=false in AJAX request', function() {
```
johnhawkinson: Can we be more expository? E.g.

```patch
diff --git a/spec/RecapSpec.js b/spec/RecapSpec.js
index e3933d0..e092f36 100644
--- a/spec/RecapSpec.js
+++ b/spec/RecapSpec.js
@@ -148,7 +148,7 @@ describe('The Recap export module', function() {
 
     var bytes = new Uint8Array([100, 100, 200, 200, 300]);
 
-    it('all send debug=false in AJAX request', function() {
+    it('all send debug=false in AJAX request.\nThis probably means debug=TRUE is set in recap.js:4.', function() {
       expected = new FormDataFake();
       expected.append('debug', false);
```

So that the result is a little more helpful:

```
pb3:recap-chrome jhawk$ karma start --single-run --browsers Chrome_CI
20 03 2018 20:39:04.703:INFO [karma]: Karma v2.0.0 server started at http://0.0.0.0:9876/
20 03 2018 20:39:04.707:INFO [launcher]: Launching browser Chrome_CI with unlimited concurrency
20 03 2018 20:39:04.711:INFO [launcher]: Starting browser Chrome
20 03 2018 20:39:05.017:INFO [HeadlessChrome 0.0.0 (Mac OS X 10.10.5)]: Connected on socket RKr45QP1n7b7urv7AAAA with id 2307793
HeadlessChrome 0.0.0 (Mac OS X 10.10.5) The Recap export module uploadDocket, uploadAttachmentMenu, uploadDocument all send debug=false in AJAX request.
This probably means debug=TRUE is set in recap.js:4. FAILED
	Expected FormDataFake({ court: 'nysd', pacer_case_id: '437098', upload_type: 1, filepath_local: [object Blob], debug: true }) to equal <jasmine.objectContaining(FormDataFake({ debug: false }))>.
	    at UserContext.<anonymous> (spec/RecapSpec.js:157:26)
	Expected FormDataFake({ court: 'nysd', pacer_case_id: '437098', upload_type: 2, filepath_local: [object Blob], debug: true }) to equal <jasmine.objectContaining(FormDataFake({ debug: false }))>.
	    at UserContext.<anonymous> (spec/RecapSpec.js:161:26)
	Expected FormDataFake({ court: 'nysd', pacer_case_id: '437098', pacer_doc_id: '127015406472', document_number: '4', attachment_number: '1', filepath_local: [object Blob], upload_type: 3, debug: true }) to equal <jasmine.objectContaining(FormDataFake({ debug: false }))>.
	    at UserContext.<anonymous> (spec/RecapSpec.js:167:26)
HeadlessChrome 0.0.0 (Mac OS X 10.10.5): Executed 93 of 93 (1 FAILED) (0.123 secs / 0.118 secs)
pb3:recap-chrome jhawk$ 
```

rather than just:

```
...
HeadlessChrome 0.0.0 (Mac OS X 10.10.5) The Recap export module uploadDocket, uploadAttachmentMenu, uploadDocument all send debug=false in AJAX request FAILED
...
```

spec/RecapSpec.js
```diff
@@ -133,6 +133,42 @@ describe('The Recap export module', function() {
...
+      expected = new FormDataFake();
+      expected.append('debug', false);
+
+      recap.uploadDocket(court, pacer_case_id, html, 'DOCKET', function() {});
+      var actualData = jasmine.Ajax.requests.mostRecent().data();
+      expect(actualData).toEqual(jasmine.objectContaining(expected));
```
johnhawkinson: I thought we didn't want to use `objectContaining()` in a case like this when strict equality. would work?

Also, should we not be changing the `objectContaining()`s from my commit to `jasmine.any(String)` or `jasmine.any(Boolean)` or whatnot?

spec/RecapSpec.js
audiodude: Done

spec/RecapSpec.js
audiodude: Done

spec/RecapSpec.js
audiodude: Right, we don't need objectContaining() here because we just want to check one value. Updated.

As far as objectContaining, there's no inherent harm in using that (except that if the methods started sending extra parameters we wouldn't catch it). We don't actually want to use jasmine.any() because that's only for cases, like the nonce you mentioned, where there's some value and all we know about it is that it is a String, we can't know the value to test for strict equality.

mlissner:
@audiodude, I'm doing some cleanup, and I'm wondering if I should just close this PR. OTOH, I don't want your work to go to waste. I honestly don't know what happened here. What do you think? Should we take another stab at this? I'm happy to do so if you think there's value to it.

audiodude:
I don't really remember what this is for, but if you've gotten by for a year and a half without this PR, you probably don't need it.

mlissner:
Ha! I misread the year. Yeah, let's move on. Sorry this didn't get merged.

PR #48, opened 2018-03-27
feat(recap): Sends the installed version to any URL in the manifest.

mlissner:
Pretty basic changes here. All we do is set up the extension to send the version string to any website where it's running and tweak the manifest to send that to a few new places:

 - https://www.courtlistener.com
 - locahost & 127.0.0.1
 - removes the `include_globs` which would normally filter the domain further (this has the side effect of making RECAP work across all of uscourts.gov instead of only on certain domains/urls).

Fixes: freelawproject/recap#215.

### DIFF
--- a/content.js
+++ b/content.js
@@ -40,3 +40,11 @@ if (PACER.hasPacerCookie(document.cookie)) {
 } else {
   console.info(`RECAP: Taking no actions because not logged in: `+url);
 }
+
+// Dispatch a message to every URL that's in the manifest to say that RECAP is
+// installed.
+let currentVersion = chrome.runtime.getManifest().version;
+window.postMessage({
+  sender: "recap-extension",
+  message: currentVersion
+}, "*");
--- a/content.js
+++ b/content.js
@@ -40,4 +40,3 @@ if (PACER.hasPacerCookie(document.cookie)) {
 } else {
   console.info(`RECAP: Taking no actions because not logged in: `+url);
 }
-
--- a/content.js
+++ b/content.js
@@ -41,10 +41,3 @@ if (PACER.hasPacerCookie(document.cookie)) {
   console.info(`RECAP: Taking no actions because not logged in: `+url);
 }
 
-// Dispatch a message to every URL that's in the manifest to say that RECAP is
-// installed.
-let currentVersion = chrome.runtime.getManifest().version;
-window.postMessage({
-  sender: "recap-extension",
-  message: currentVersion
-}, "*");
--- a/install_notifier.js
+++ b/install_notifier.js
@@ -1,7 +1,12 @@
 // Dispatch a message to every URL that's in the manifest to say that RECAP is
-// installed.
-let currentVersion = chrome.runtime.getManifest().version;
-window.postMessage({
-  sender: "recap-extension",
-  message: currentVersion
-}, "*");
+// installed.  This allows webpages to take action based on the presence of the
+// extension and its version. This is only allowed for a small whitelist of
+// domains defined in the manifest.
+(function () {
+  let currentVersion = chrome.runtime.getManifest().version;
+  window.postMessage({
+    sender: "recap-extension",
+    message_name: "version",
+    message: currentVersion
+  }, "*");
+})();
--- a/install_notifier.js
+++ b/install_notifier.js
@@ -1,0 +1,7 @@
+// Dispatch a message to every URL that's in the manifest to say that RECAP is
+// installed.
+let currentVersion = chrome.runtime.getManifest().version;
+window.postMessage({
+  sender: "recap-extension",
+  message: currentVersion
+}, "*");
--- a/manifest.json
+++ b/manifest.json
@@ -38,9 +38,7 @@
     ]
   },
   "content_scripts": [{
-    "matches": [
-      "*://*.uscourts.gov/*"
-    ],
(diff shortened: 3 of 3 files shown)


content.js
```diff
@@ -40,3 +40,4 @@ if (PACER.hasPacerCookie(document.cookie)) {
 } else {
   console.info(`RECAP: Taking no actions because not logged in: `+url);
 }
+
```
johnhawkinson: Spurrious newline.

install_notifier.js
```diff
@@ -0,0 +1,7 @@
+// Dispatch a message to every URL that's in the manifest to say that RECAP is
+// installed.
+let currentVersion = chrome.runtime.getManifest().version;
```
johnhawkinson: This makes me uncomfortable, but I'm not quite sure if I'm right — I've still not entirely wrapped my head around `let`. But as I understand it, `let` at the top of a file is not guaranteed to be file scoped. https://esdiscuss.org/topic/let-and-file-scope suggests that top-level `let`s in a file may be global-scoped, unless this is one of the exceptions.

So I'd feel more comfortable if this was wrapped in a closure or a block.
(Also, I'm not sure you need a variable there. You could just move the `getManifest()` call down 3 lines.)

install_notifier.js
```diff
@@ -0,0 +1,7 @@
+// Dispatch a message to every URL that's in the manifest to say that RECAP is
+// installed.
```
johnhawkinson: Perhaps a little more clarity of purpose:

```patch
diff --git a/install_notifier.js b/install_notifier.js
index 2bac7dd..e2225e3 100644
--- a/install_notifier.js
+++ b/install_notifier.js
@@ -1,2 +1,4 @@
-// Dispatch a message to every URL that's in the manifest to say that RECAP is
-// installed.
+// Dispatch a message to every URL in the manifest for this file
+// to provide the RECAP version. This allows webpages to take
+// action based on the presence of the extension and its version.
+// This is only allowed for a small whitelist of domains.
 ```

File: install_notifier.js [hunk 4]
```diff
@@ -0,0 +1,7 @@
...
+let currentVersion = chrome.runtime.getManifest().version;
+window.postMessage({
+  sender: "recap-extension",
+  message: currentVersion
```
johnhawkinson: It seems unlikely this is the only message we will ever send. Since `window.postMessage()` automatically serializes objects, and indeed you're making use of that by sending an object with `sender` and `message` properties (rather than just a bare string), it seems like you should include a type, so that other messages can be disambiguated in the future.

I.e.:

```js
{
  sender: "recap-extension",
  type: "extension-version",
  message: currentVersion,
}
```
I suppose you could argue this is unnecessary since you've already added the nonstandard `sender` property which allows disambiguation, but I would still include more. Or rename `sender` to `type` and make its value `recap-extension-version`. Or something.

manifest.json
```diff
@@ -58,6 +58,12 @@
       "content.js"
     ],
     "run_at": "document_end"
+  }, {
+    "matches": ["https://www.courtlistener.com/*"],
+    "js": [
+      "install_notifier.js"
+    ],
+    "run_at": "document_end"
```
johnhawkinson: Use of `document_end` rather than the default `document_idle` opens you up to potential race conditions with scripts for the page that are still executing. Presumably you're aware of this and confident that this is the right choice, or prepared to adjust it if not. Cf. https://developer.mozilla.org/en-US/Add-ons/WebExtensions/manifest.json/content_scripts.

content.js
mlissner: This is now part of the editor config. They'll be appearing...

johnhawkinson: That's unfortunate, it screws up all kinds of `git blame` operations. If you want to do it, please make sure it's a separate commit that doesn't contain any functional changes (not squashed with or integrated with the rest).

Or go through and touch all files at once. (Disfavored. But not as much as having it happen randomly.)

But this is the reason that such editor configuration automatic linting/KNFing/whatever are generally disfavored. They produce more churn in the code that makes it harder for other developers and future people trying to find the history of a change.

But also:
Why is this happening on this file that is not otherwise being touched in this changset?
Why would you want to have an EXTRA blank line at the end of a file. It's a reasonable style rule to have make sure that files end in a final newline (cf. `require-final-newline t`), but not a common one to require they end in **two**.

install_notifier.js
mlissner: Tweaked to be a closure. Good call.

install_notifier.js
mlissner: Sure. Tweaked.

install_notifier.js
mlissner: I made this into:

```
    window.postMessage({
      sender: "recap-extension",
      message_name: "version",
      message: currentVersion
    }, "*");
```

So, `sender` lets us figure out what's sending the message (the RECAP extension), and then `message_name` gives us the variable name.

manifest.json
mlissner: No, actually, I assumed that `document_end` was the normal case, since it's what we normally use. Thanks, tweaked too. 

Jeesh. Small PR, but many tweaks.

content.js
mlissner: > Why is this happening on this file that is not otherwise being touched in this changset?

Only because I had monkeyed with it at first, then didn't catch that I left extra newlines on there. So this one is on me.

PR #54, opened 2019-02-23
Tidyup

jraller:


johnhawkinson:
Err…since when do we just commit PRs without waiting any time for review (and on a weekend at that)? (Isn't the whole point of using a PR to facilitate review?) I suppose I'm unclear on our local practice, really.

I would have said:

* I'm really not sure it's a good idea to do this kind of mass refactoring that generally provides no actual benefit (e.g. `var -> let` changes). It means anybody with a dev tree with changes is going to have annoying merge problems, and it's not as if we are going to see any code quality improvement or bugfixes as a result. 

* Where these commits are not squashed, which they weren't when they were committed, their commit messages aren't adequate. Commit messages should be clear on scope. Generally in git that means including the filename, module name, or function name as the prefix to the commit message. If they're going to be squashed, well, not such a big deal. The fact that there are five commits here with identical messages is a sign of a problem.

Less prominently:

* Functional changes are mixed in with mass refactoring changes. That is not good. E.g., f37a987 changes the CSS, which could have an actual effect on something; 72d129 changes html, with the same ramifications. But 21cc59f is a `var -> let` change, which is just a style change and shouldn't be mixed in. (Of course, since the changes weren't squashed, this isn't a huge deal. If they're not squashed, then it comes down to how easy it is to facilitate review.)
 
* 611ebd2 rewrites style but ignores an xxx comment about the style. Presumably the comment should go. Sign of review issues.

* fe8ef29 and 1edef7 are "refactored assignment to match out of conditional statements" but that is not what they do.  Instead they are whitespace and semicolon changes.

* cc167da changed a top-level `var` to a `let`, for `DEBUGLEVEL`, which seems like it could have undesirable repercussions, but maybe I should think a little harder about that.

**Because this has only landed for less than a day, I would force-push back the tree to revoke these commits and review them in the PR process and commit them with proper commit messages. If I were King.**

PR #55, opened 2019-02-23
fix(pacer.js): Fixing an error I introduced that treated caseId and c…

jraller:


johnhawkinson:
(Also, the line 238 comment should be addressed. Probably at the same time, although you could argue it could be a seperate PR. One PR per line, w00t!)

johnhawkinson:
What's up with the indentation mismatch on 244/245/246?

PR #58, opened 2019-03-10
refactor(src): move src files to src directory

jraller:
update readme zip command, update karma.conf.js


mlissner:
I had a comment somewhere (a previous PR that did this?) that expressed some concern about how this might break other PRs or branches that we have outstanding. I *think* this should be generally OK and that git should handle this properly since everything is marked as a file move, so I went ahead and pulled this. I also imagine you're pretty eager to get to meatier things.

PR #60, opened 2019-03-14
test(karma.conf.js, package.json): update to latest version of karma …

jraller:
…and jasmine, fix random test execution issue

The upgrade crossed a breaking change in jasmine where by default tests are run in random order now. This is a desirable thing in the future, but would require refactoring at least three of the groups of tests, so for now a setting has been added to keep the previous behavior of running tests in the order in which they appear in the file.


mlissner:
Good god, `package-lock.json` is...something to behold.

PR #61, opened 2019-03-15
test(PacerSpec.js): 100% code coverage, tests in order, fixture cleanup

jraller:
Added tests for getDocumentIdFromForm and getCaseNumberFromInputs as primary task. Reordered other test groups to match order of the functions in pacer.js. Added additional explicit tests for things that already had some code coverage through other testing.

Changed from removing the innerHTML of some fixtures to removing them completely. The removal will aide in preparations for being able to execute tests in random order as it doesn't leave the testing document cluttered with the setups for tests that have already been run. Cleanup of the other two Spec files will be needed in this regard.


mlissner:
Looks great to me. 100% — hot damn.

PR #62, opened 2019-03-16
test(ContentDelegateSpec.js): tests run without leaving behind items …

jraller:
…in the document, can be run in random order now


mlissner:
I think in the future we should do style/cleanup stuff in separate PRs if that's possible. Makes review really difficult. I just kind of accepted this one because figuring out the actually changes was...difficult and I expect you're doing good things.

PR #63, opened 2019-03-17
build(package.json): added existing dependencies to node_modules, no …

jraller:
…change to src/assets/js.

no diff for jquery and momentjs, minor diffs for file-saver and
livestamp.

file-saver reports save version number, our existing version
had view.safari

livestamp is version 2.0.0 compared to our 1.3.2 has
new code for module, and now deletes prior object on update

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -773,6 +773,11 @@
       "integrity": "sha1-PYpcZog6FqMMqGQ+hR8Zuqd5eRc=",
       "dev": true
     },
+    "file-saver": {
+      "version": "1.3.2",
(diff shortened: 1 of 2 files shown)


mlissner:
Are there any changes here that we should note in the readme?

jraller:
Right now the changes are only informational. We are acknowledging the packages we are using, and we can get security notifications from GitHub based on what is listed in the package.json. I did not add any script or package at this point for moving those to the src/assets/js folder. Once that is done I agree that we should add instructions to the README.md file. Updating the versions of those files in the package.json at this point does not have an effect on the extension.

mlissner:
Makes sense. Thanks for the explanation.

PR #64, opened 2019-03-19
refactor(pacer.js): avoid fallthrough in switch, move assignment out …

jraller:
…of comparison

Grouped some cleanup together.

### DIFF
--- a/src/pacer.js
+++ b/src/pacer.js
@@ -121,6 +121,7 @@ let PACER = {
 
         default:
           debug(4, `Assuming servlet ${servlet} is not a docket.`);
+          return false;
         case 'CaseSearch.jsp':
         case 'ShowDoc':
         case 'ShowDocMulti':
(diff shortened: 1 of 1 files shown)


src/pacer.js
```diff
@@ -121,6 +121,7 @@ let PACER = {
 
         default:
           debug(4, `Assuming servlet ${servlet} is not a docket.`);
+          return false;
```
johnhawkinson: If you're going to eliminate the fallthrough, then you also need to add a newline before the next group of `case`s, since they are blocked by comon fallthrough. i.e. on line 125.

src/pacer.js
```diff
@@ -236,14 +237,16 @@ let PACER = {
           }
         }
         // xxx does not match style above.
```
johnhawkinson: This `xxx` comment is no longer apposite.

src/pacer.js
```diff
@@ -236,14 +237,16 @@ let PACER = {
...
-        let match;
```
johnhawkinson: Since you now assign to `match` twice, it seems like you'd be better off manually hoisting the `let match` up so that both assignments to `match` (at 240 and 246) looked similar, rather than having a `let` on the first one. YMMV I suppose?

jraller: combining all three match usages under a single let match;

src/pacer.js
jraller: Removing comment.

src/pacer.js
jraller: Adding blank line.

PR #65, opened 2019-03-20
fix(options.html): fix accessibility

jraller:
set language
attached labels to inputs in unused section
added alt tags to images

### DIFF
--- a/src/options.html
+++ b/src/options.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html>
+<html lang="en">
 <head>
   <meta charset="utf-8"/>
   <title>RECAP Options</title>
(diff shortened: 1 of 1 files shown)


johnhawkinson:
Please put whitespace changes and typographical fixes inside comments in commits separate from functional changes.

mlissner:
I think these kinds of really minor changes are fine, @johnhawkinson. You're right technically, but so long as the PR is readable, I'm happy.

johnhawkinson:
The problem here is not the PR, Mike. The problem is the horrible pain it causes in the future when you try to use `git annotate` (`git blame`) and have to dscend through whitespace changes and discover that you cannot differentiate whitespace commits from actual bug-introducing commits. Huge pain.

Never do this. No exceptions. You will thank me later.

PR #66, opened 2019-04-17
Cleanup

jraller:
Test coverage increased.


mlissner:
This is too much to review all at once and fully understand by looking at the diff, but it only touches tests, where I know you've been working hard, so risk here is low of a bug that causes any issues and I'm pretty confident it's solid progress. Sorry if I dropped this one. I didn't realize it was outstanding.

PR #73, opened 2019-08-03
FIX for RECAP fails to load documents for Northern District of Georgia #277

lkj19zpdiDDe:
Fix for https://github.com/freelawproject/recap/issues/277

### DIFF
--- a/src/content_delegate.js
+++ b/src/content_delegate.js
@@ -352,7 +352,7 @@ ContentDelegate.prototype.onDocumentViewSubmit = function (event) {
     // in it, which is handled by showPdfPage.
     if (type === 'application/pdf') {
       // canb and ca9 return PDFs and trigger this code path.
-      var html = '<style>body { margin: 0; } iframe { border: none; }' +
+      let html = '<style>body { margin: 0; } iframe { border: none; }' +
                  '</style><iframe src="' + URL.createObjectURL(blob) +
                  '" width="100%" height="100%"></iframe>';
       this.showPdfPage(document.documentElement, html, previousPageHtml,
@@ -366,15 +366,15 @@ ContentDelegate.prototype.onDocumentViewSubmit = function (event) {
           // this was first display by the Northern District of Georgia
           // https://github.com/freelawproject/recap/issues/277
           const redirectResult = Array.from(html.matchAll(/(?:window\.location\s=\s")(.*)(?:\")/g));
-          if(redirectResult && redirectResult.length>0) {
-            let url = redirectResult[0][1];
+          if (redirectResult.length > 0) {
+            const url = redirectResult[0][1];
             html = '<style>body { margin: 0; } iframe { border: none; }' +
-                 '</style><iframe src="' + url +
-                 '" width="100%" height="100%"></iframe>';
+                   '</style><iframe src="' + url +
+                   '" width="100%" height="100%"></iframe>';
           }
           this.showPdfPage(
-              document.documentElement, html, previousPageHtml,
-              document_number, attachment_number, docket_number);
+            document.documentElement, html, previousPageHtml,
+            document_number, attachment_number, docket_number);
       }.bind(this);
       reader.readAsText(blob);  // convert blob to HTML text
     }
--- a/src/content_delegate.js
+++ b/src/content_delegate.js
@@ -361,9 +361,20 @@ ContentDelegate.prototype.onDocumentViewSubmit = function (event) {
       // dcd (and presumably others) trigger this code path.
       var reader = new FileReader();
       reader.onload = function() {
+          let html=reader.result;
+          // check if we have an HTML page which redirects the user to the PDF
+          // this was first display by the Northern District of Georgia
+          // https://github.com/freelawproject/recap/issues/277
+          const redirectResult = Array.from(html.matchAll(/(?:window\.location\s=\s")(.*)(?:\")/g));
+          if(redirectResult && redirectResult.length>0) {
+            let url = redirectResult[0][1];
+            html = '<style>body { margin: 0; } iframe { border: none; }' +
+                 '</style><iframe src="' + url +
+                 '" width="100%" height="100%"></iframe>';
+          }
           this.showPdfPage(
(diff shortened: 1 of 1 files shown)


src/content_delegate.js
```diff
@@ -361,9 +361,20 @@ ContentDelegate.prototype.onDocumentViewSubmit = function (event) {
       // dcd (and presumably others) trigger this code path.
       var reader = new FileReader();
       reader.onload = function() {
+          let html=reader.result;
+          // check if we have an HTML page which redirects the user to the PDF
+          // this was first display by the Northern District of Georgia
+          // https://github.com/freelawproject/recap/issues/277
+          const redirectResult = Array.from(html.matchAll(/(?:window\.location\s=\s")(.*)(?:\")/g));
```
johnhawkinson: The last time I had to do something similar, I used
`      if (match = /window\.location\s*=\s*["']([^"']+)["'];?/.exec(script.text))`
In particular:
. Why is the first non-capturing group present? It does not appear to serve any grouping function.
. Whitespace need not appear around the equals sign, so this is more robust with `\s*` rather than `\s`; and it could be more than one character long.
. JS strings need not be delimited by double-quotes, although my regexp may do the wrong thing if there is single-quoted string inside a double-quoted string, so I don't claim it's correct. (I don't think JS regexps have a great way to do this without two passes, unlike Python regexps with named groups  (`(?P=name)`). I guess you can do two passes if this were really important.)
. The final non-capturing group doesn't seem to serve a grouping function either. "When everything is a group, nothing is" — THE INCREDIBLES.
. What does the `/g` flag do on a search that is not a replacement?

src/content_delegate.js
```diff
@@ -361,9 +361,20 @@ ContentDelegate.prototype.onDocumentViewSubmit = function (event) {
...
+          if(redirectResult && redirectResult.length>0) {
+            let url = redirectResult[0][1];
+            html = '<style>body { margin: 0; } iframe { border: none; }' +
+                 '</style><iframe src="' + url +
+                 '" width="100%" height="100%"></iframe>';
+          }
```
johnhawkinson: Maybe it's just me, but it's weird to close the style tag and open the iframe tag on the same line, when the whole thing is multiline. Perhaps a template literal? I 'm not up on my CSS enough to know if all of these explicit specifications are necessary? Also can't we `<iframe/>`? Maybe not.

On the other hand, I guess this is what the existing code does above, so I suppose I should shut up about it.

The `let`/`var` police are probably going to show up and complain about treating `html` as if it were a block-scoped variable and reusing it from the prior block, because the variable declaration doesn't match the control flow, although it works fine because it is hoisted. But perhaps those cops should be appeased.

src/content_delegate.js
```diff
@@ -361,9 +361,20 @@ ContentDelegate.prototype.onDocumentViewSubmit = function (event) {
...
           this.showPdfPage(
-            document.documentElement, reader.result, previousPageHtml,
-            document_number, attachment_number, docket_number);
+              document.documentElement, html, previousPageHtml,
+              document_number, attachment_number, docket_number);
```
johnhawkinson: Why is there a whitespace change here? And of course, under ordinary circumstances whitespace changes should be a separate commit, but maybe that's too picky here.

src/content_delegate.js
lkj19zpdiDDe: Yeah, exactly. I was just matching the style on line 355 vs. putting my own style in. (I'd prefer to use template literals too.)  Bigger picture, I didn't spot a style guide or any linting. Let me know if you have any interest in those, and I'm happy to make a separate PR/contribution of those.

lkj19zpdiDDe: On the scoping of `HTML`, I converted the declaration on line 355 to be a `let` making them both scoped to the block.

src/content_delegate.js
lkj19zpdiDDe: A mistake on this one. I've removed the extra space.

src/content_delegate.js
```diff
@@ -361,9 +361,20 @@ ContentDelegate.prototype.onDocumentViewSubmit = function (event) {
       // dcd (and presumably others) trigger this code path.
       var reader = new FileReader();
       reader.onload = function() {
+          let html=reader.result;
+          // check if we have an HTML page which redirects the user to the PDF
+          // this was first display by the Northern District of Georgia
+          // https://github.com/freelawproject/recap/issues/277
+          const redirectResult = Array.from(html.matchAll(/(?:window\.location\s=\s")(.*)(?:\")/g));
+          if(redirectResult && redirectResult.length>0) {
```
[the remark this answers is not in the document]
lkj19zpdiDDe: Moved to only check the length.

src/content_delegate.js
lkj19zpdiDDe: Adjusted along the lines of what's suggested.

On the /g flag, it finds all matches vs. stopping at a single match. Generally, /g is used with matchAll, but, in this particular case, we're not handling multiple results.

johnhawkinson: Right. I guess I meant "Why is the /g appropriate here?" Seems like it's not?

src/content_delegate.js
johnhawkinson: Feel free to swap the prior one to a template literal in a separate commit and rewrite history to have it come first. Or could be a separate PR if you want.

Style is...complicated but there is not a clear and rigorous guide here, so.

mlissner:
Wow, thanks for all the activity here. I've skimmed through this, but I think @knowtheory is going to do another round of review. @knowtheory, if you wan to tune this up (if it still needs it), and do a final review, I'll mash the pull button.

@lkj19zpdiDDe, if you wanted to do a first pass at a style guide, I'd welcome it, but it's probably worth starting a discussion in an issue first, so we can figure out the big picture before you dial into the details? I'm very much +1 on style guides, because they help in situations like this.

mlissner:
This was a wonderful team effort. Thank you all for chipping in. I'll get a release tested and out the door.

PR #80, opened 2019-08-06
Cleanup PR

knowtheory:
Hey folks,

This is the start of a cleanup branch.  gonna get some JS linting configs up in here.

Thoughts/feelings/etc would be appreciated!  (the obvious changes here are replacing string concat with template strings and other typo-proofing)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,39 @@
+module.exports = {
+    'env': {
+        'browser': true,
+        'es6': true
+    },
+    //'extends': 'eslint:recommended',
+    'globals': {
+        'Atomics': 'readonly',
+        'SharedArrayBuffer': 'readonly'
+    },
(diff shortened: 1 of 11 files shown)


knowtheory:
Okay, i just added an eslint config.  The repo does not currently pass this eslint config, but i wanted to get one in so that folks can see what it looks like and what it'll do.

This isn't entirely where i was hoping to spend my time, but i think it's worth getting some ground rules in.  I'm also, generally speaking, not a prescriptivist, nor a fan of bikeshedding style rules :)

All of that said, I think there are a few things that tools like eslint can help catch and prevent other people from having trouble figuring out how best to contribute.

Particularly rules like "don't use `var`", "strings should all be single quoted", "use template strings instead of string concatenation", are all helpful for ensuring that bugs can be preemptively stomped.

PR #83, opened 2019-08-13
Fix donate link

knowtheory:
The RECAP and Donate links in the extension currently don't open anything in Google Chrome (despite working fine in Firefox).  Adding `target="_blank"` is required for Chrome extensions to open new windows.


mlissner:
Thanks. I'm about 95% sure I tested this in both Chrome and FF when I made those links. Grrr.

PR #87, opened 2019-08-23
Add a search form to the options dialog

knowtheory:
Okay!  I've added two headings and a fieldset with a search form.

There are a bunch of design questions, one of which is a blocker for merging this.

In order to visually differentiate the config options for the plugin from the other elements i've added two headers (just using the default browser styles like most of the rest of the plugin).

Unfortunately, Firefox addon popovers have [max dimensions of 800px by 600px](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/user_interface/Popups).  This results in the twitter, newsletter, & fb icons as well as the version number disappearing below the fold and requiring scrolling to find.

I've tried a couple variations to get something that isn't major surgery, but at this point i think that some rearranging is necessary to 1) ensure all the options are present & clear and 2) that the configuration options are visually separated from the search form, attribution & other buttons.

So, i'm submitting this for feedback/thoughts!


knowtheory:
One possible avenue is to pull more of the CourtListener bootstrap styles into the plugin (also because there are visual differences in the default styles for Firefox and Chrome plugins)

mlissner:
OK, so following some slack conversation, I think it'd be great to try to redesign the popup to be a little more sophisticated:

1. Introduce tabs. One for "RECAP Archive" and the other for "Settings"
2. Split off the "disable RECAP" functionality into a toggle that is pulled out of the settings. Maybe use something like: http://www.bootstraptoggle.com/

Some additional considerations:

1. The options.html and the popup stuff is currently...the same, I think. Can we split these off or does it make sense to do so?
2. Making this look better probably means using Bootstrap. That's fine by me. A little more JS and whatnot in an extension is fine in my book.
3. By pulling search into its own tab, we should be able to make it a more sophisticated interface. For example, it could have links to advanced search, a semi-advanced search interface, search tips, etc.

A sketch would be a great place to start with this.

knowtheory:
Yep, i like this direction.

re 3: the `manifest.json` has configuration points for the `default_popup` and an `options`.  We currently have both pointed to `options.html`.  So it should be possible just to split those up to point to two different places.

knowtheory:
1. Yep, i agree.  I think "enable uploads" is probably my favorite of those bullets and probably correct.
2. Yeahhhhh i've been puzzling over what to do there.  My general disposition is very hesitant to remove the big DONATE call to action, but i'd rather make that predicated on actual data, like if anyone ever clicks through on the current call to action.  Ideally this should fit in w/ your overall donation strategy & funnel.  A donate form is an interesting way to go.
3. Yep!  I think you're right on that, I'm also interested generally on the style unification w/ Courtlistener.  Do you have a preference there?  Making them smaller & bold seems like a good way to go and will help open up some whitespace (which the design needs more of).

mlissner:
1. Great.

2. We don't get many donations that way. Once or twice a year we open a tab for people to donate (totally interrupting their workflow), but otherwise, looking at logs, it doesn't look like that link drives much, so I feel OK demoting it. It probably does serve as good advertising though. Eh. I think it can go for the sake of polish.

3. Unification with CL is nice, but not essential. I think I'd keep my eye on that so things don't get way out there, but I'd just focus on making it look good on its own for the moment. So small & bold is probably perfectly good.

knowtheory:
💪 No panes, no gains 💪

Here are more screenshots of how things look 😃 

Screen Shot 2019-09-19 at 7 38 47 PM
Screen Shot 2019-09-19 at 7 38 53 PM
Screen Shot 2019-09-19 at 8 32 23 PM

PR #110, opened 2020-01-24
fix 235

litewarp:
Chrome reports that large PDFs exceed the permissible length in messages between the background worker and the content script. 

```
Error handling response: TypeError: Error in invocation of runtime.sendMessage(optional string extensionId, any message, optional object options, optional function responseCallback): Message length exceeded maximum allowed length.
    at Object.sender.<computed> [as uploadDocument] (chrome-extension://dndpdemfjohfjfnaahiaoknbccpghcmb/utils.js:77:24)
    at ContentDelegate.<anonymous> (chrome-extension://dndpdemfjohfjfnaahiaoknbccpghcmb/content_delegate.js:576:26)
```

This PR fixes the problem by utilizing the DOM to store the blob and passing a local URL instead of the actual blob itself to the background worker. Tested on the two links in issue 235, which now have correctly uploaded documents.

### DIFF
--- a/src/content_delegate.js
+++ b/src/content_delegate.js
@@ -415,8 +415,8 @@ ContentDelegate.prototype.showPdfPage = function(
     history.replaceState({content: previousPageHtml}, '');
 
     // store the result in chrome local storage
-    //
-    const nonce = "EB23C5DF"
+
+    const nonce = "blob_upload_storage"
     const data = { [nonce]: ab }
 
     chrome.storage.local.set(data, function() {
--- a/src/content_delegate.js
+++ b/src/content_delegate.js
@@ -414,10 +414,20 @@ ContentDelegate.prototype.showPdfPage = function(
     };
     history.replaceState({content: previousPageHtml}, '');
 
+    // store the result in chrome local storage
+    //
+    const nonce = "EB23C5DF"
+    const data = { [nonce]: ab }
+
+    chrome.storage.local.set(data, function() {
+      console.log(`Blob saved: ${data.nonce.length}`)
+    })
+
     // Get the PACER case ID and, on completion, define displayPDF()
     // to either display the PDF in the provided <iframe>, or, if
     // external_pdf is set, save it using FileSaver.js's saveAs().
     let blob = new Blob([new Uint8Array(ab)], {type: type});
+
     this.recap.getPacerCaseIdFromPacerDocId(
       this.pacer_doc_id, function(pacer_case_id){
         console.info(`RECAP: Stored pacer_case_id is ${pacer_case_id}`);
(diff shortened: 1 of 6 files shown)


src/content_delegate.js
```diff
@@ -482,14 +482,11 @@ ContentDelegate.prototype.showPdfPage = function(
               }
             }.bind(this);
 
-            // In Chrome, blobs can't be passed from content scripts
-            // to background scripts, so we have to convert to an
-            // array and pass that, then convert back to a blob when
-            // we add the data to the FormData object.
-            let bytes = arrayBufferToArray(ab);
```
mlissner: Can you also remove the arrayBufferToArray function? I think it's no longer used.

litewarp: Done.

mlissner:
I come with bad news. Firefox has no love for this approach, and reports:

> 22:36:05.679 Security Error: Content at moz-extension://c8d95e3b-c2b2-417a-b7af-d3eb5abc5d27/_generated_background_page.html may not load data from blob:https://ecf.ned.uscourts.gov/674d67b2-e1bf-4555-b2b8-b07b7938ae1c.
> 22:36:05.679 TypeError: NetworkError when attempting to fetch resource.

Among several other errors that I think might be new. I guess this confirms my suspicion that Chrome is doing a weird thing here by allowing background scripts to access this in the content DOM. Firefox isn't playing that game. Bummer.

src/content_delegate.js
```diff
@@ -414,10 +414,20 @@ ContentDelegate.prototype.showPdfPage = function(
     };
     history.replaceState({content: previousPageHtml}, '');
 
+    // store the result in chrome local storage
+
+    const nonce = "blob_upload_storage"
+    const data = { [nonce]: ab }
```
mlissner: I'm a little perplexed at two things:

1. Maybe this is a JS thing, but why do you pass `[nonce]` as the key instead of just using `nonce`? 

2. `nonce` feels like the wrong variable name. Maybe `file_storage_key` or something like that?

litewarp: I'm agnostic w/r/t name - I started with a nonce when I thought about each blob being stored with a unique key, but then moved to the fixed string key. 

In JS you normally couldn't set a key dynamically simply by passing in the variable name (I believe it's because if you use `{nonce: data }` the key will be `nonce` and not whatever the variable is). Now with ES6, you can use computed property names by enclosing the variable key in brackets. See "new notations" section of the [Object Initializer Docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer)

litewarp: Another comment on 1 - you could remove the variable all-together since it's a fixed string, but the only downside to that is that we currently pass the nonce (or whatever its called) to the uploadDocument function. So all the tests/related functions would need to be altered ... figured we could keep the variable, pass it, and then have the infrastructure to use actual dynamic keys later.

mlissner: Got it. Merging, but with the understanding that we may have more testing to do.

PR #113, opened 2020-02-13
fix 117 and 235

litewarp:
1. Pass PDF and Zip files as array of bytes through chrome local storage

2. Add functions to detect if download zip file page, and to intercept the zip download and upload to recap. 

Still to do -- confirm upload with endpoint, redirect zip file to client if upload callback is ok (or maybe even if not).


src/content_delegate.js
```diff
@@ -585,31 +588,186 @@ ContentDelegate.prototype.attachRecapLinkToEligibleDocs = function() {
...
+
+  // helper function - convert string to html document
+  const stringToDocBody = (str) => {
+    const parser = new DOMParser();
+    const newDoc = parser.parseFromString(str, 'text/html')
+    return newDoc.body
+  }
+
+  // helper function - returns filename based on user preferences
+  const generateFileName = (options, pacerCaseId) => {
```
litewarp: @mlissner here's the filename function

src/recap.js
```diff
@@ -169,18 +170,54 @@ function Recap() {
...
 
       });
+    },
+    // Upload a zip file to the RECAP server, calling the cb with ok flag
+    uploadZipFile: async function (pacer_court, pacer_case_id, nonce, cb) {
+      console.info(`RECAP: Attempting PDF upload to RECAP Archive with details: ` +
+        `pacer_court: ${pacer_court}, pacer_case_id: ${pacer_case_id}`)
+
+      // async call for the blob
+      const bytes = await getItemsFromStorage([nonce, 'docId'])
```
mlissner: `bytes` seems like a funny variable name here, no?

src/recap.js
```diff
@@ -169,18 +170,54 @@ function Recap() {
...
+      formData.append('debug', DEBUG);
+      formData.append('filepath_local', blob);
+
+      const fetchOptions = {
+        method: 'POST',
+        body: formData,
+        headers: {
+          'Authorization': `Token ${RECAP_TOKEN}`
+        }
+      }
```
mlissner: Presumably we'll be using this kind of a pattern more and more as we move away from jQuery. Is this worth abstracting out? Seems boilerplate-ish.

src/recap.js
```diff
@@ -169,18 +170,54 @@ function Recap() {
...
+
+      fetch(`${SERVER_ROOT}recap/`, fetchOptions)
+        .then(res => res.json())
+        .then(result => {
+          console.log("RECAP: Zip file uploaded successfully")
+          cb(result || null)
+        })
+        .catch(error => console.log(`RECAP: The following error occurred: ${error}`))
```
mlissner: How about:

```suggestion
        .catch(error => console.log(`RECAP: Error during zip upload: ${error}`))
```

src/pacer.js
```diff
@@ -160,6 +160,16 @@ let PACER = {
...
+  // Returns true if this is a "Download Documents" page (confirmation of
+  // pricing for all documents to receive a zip file with all of them)
+  isDownloadAllDocumentsPage: function(url, document) {
+    let inputs = document.getElementsByTagName("input");
+    let pageCheck =
+      !!url.match(/\/show_multidocs\.pl\?/) &&
+      inputs.length &&
+      inputs[inputs.length-1].value === "Download Documents"
+    return !!pageCheck
+  },
```
mlissner: Simple but great, thanks.

spec/ContentDelegateSpec.js
```diff
@@ -760,6 +766,44 @@ describe('The ContentDelegate class', function() {
...
+
+    it('checks options to see if recap is enabled', function() {});
+
+    it('uploads the Zip file to RECAP', function() {});
+
+    it('redirects the user to the download page and forwards the zip file', function() {})
+
+    it('calls the notifier once the upload finishes', function() {
+      expect(cd.notifier.showUpload).toHaveBeenCalled();
+    });
```
mlissner: I presume this is still on your todo list, but highlighting anyway so it gets on mine. If you want to do a different testing approach because Jasmine is old or whatever, I'm entirely open to that.

src/content_delegate.js
```diff
@@ -585,31 +588,193 @@ ContentDelegate.prototype.attachRecapLinkToEligibleDocs = function() {
...
-      });
-      recap_link.click($.proxy(this.handleRecapLinkClick, this, window, href));
-      recap_link.append($('<img/>').attr({
-        src: chrome.extension.getURL('assets/images/icon-16.png')
-      }));
-      recap_link.insertAfter(this.links[i]);
+    }, this));
+};
+
+// TODO: implement appellate court check
```
mlissner: Flagging your flag. :)

src/content_delegate.js
```diff
@@ -585,31 +588,193 @@ ContentDelegate.prototype.attachRecapLinkToEligibleDocs = function() {
...
+      // 4th from bottom
+      const matchedRow = firstTableRows[firstTableRows.length - 4];
+      const cells = matchedRow.querySelectorAll('td');
+      const document_number = cells[0].innerText.match(/\d+(?=\-)/)[0];
+      const docket_number = cells[1].innerText;
+      return [
+        PACER.COURT_ABBREVS[this.court],
+        docket_number,
+        document_number,
+      ].join('_').concat('.zip');
```
mlissner: There's a lot of fragile stuff here. Is it worth putting in a try with a simple file name if something goes wrong (like you do for the ia_style_filename branch?

src/content_delegate.js
```diff
@@ -585,31 +588,193 @@ ContentDelegate.prototype.attachRecapLinkToEligibleDocs = function() {
...
+  history.replaceState({content: document.documentElement.innerHTML}, '');
+
+  try {
+    // runtime start
+    $("body").css("cursor", "wait");
+    // fetch the html page which contains the <iframe> link to the zip document.
+
+    // in Firefox, use content.fetch for content-specific fetch requests
+    // https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Content_scripts#XHR_and_Fetch
+    const browserSpecificFetch = (navigator.userAgent.indexOf('Chrome') < 0) ? content.fetch : window.fetch
```
mlissner: Should this be pulled out and generalized somewhere for future intrepid developers to not feel crazy?

src/content_delegate.js
```diff
@@ -585,31 +588,193 @@ ContentDelegate.prototype.attachRecapLinkToEligibleDocs = function() {
...
+        return new Blob([buffer], {type: 'application/zip'});
+      })
+    const blobUrl = URL.createObjectURL(zipFile);
+    const pacerCaseId = (event.data.id).match(/caseid\=\d*/)[0].replace(/caseid\=/, "");
+
+    // load options
+    const payload = await getItemsFromStorage(['options']);
+    const options = payload['options'];
+
+    if (options['recap_enabled'] && !this.restricted) {
```
mlissner: Can we flip this logic to make it an early abort from the function and de-indent some of this?

This is probably bad JS, but:
```suggestion
if (!options['recap_enabled'] || this.restricted) {
  return;
}

```

src/content_delegate.js
```diff
@@ -585,31 +588,193 @@ ContentDelegate.prototype.attachRecapLinkToEligibleDocs = function() {
...
+              `<a id="recap-download" href=${blobUrl} download=${filename} width="0" height="0"/>`;
+            const htmlBody = stringToDocBody(htmlPage);
+            const frame = htmlBody.querySelector('iframe');
+            frame.insertAdjacentHTML('beforebegin', link);
+            frame.src = "";
+            frame.onload = () => document.getElementById('recap-download').click();
+            document.body = htmlBody;
+            history.pushState({content: document.body.innerHTML}, '');
+            // show notifier
+            this.notifier.showUpload('Zip uploaded to the Public Recap Archive', () => {});
```
mlissner: Nit:

```suggestion
            this.notifier.showUpload('Zip uploaded to the public RECAP Archive', () => {});
```

src/content_delegate.js
```diff
@@ -585,31 +588,193 @@ ContentDelegate.prototype.attachRecapLinkToEligibleDocs = function() {
...
+    // load options
+    const payload = await getItemsFromStorage(['options']);
+    const options = payload['options'];
+
+    if (options['recap_enabled'] && !this.restricted) {
+      this.recap.uploadZipFile(
+        this.court, // string
+        pacerCaseId, // string
+        nonce, // string
+        (ok) => { // callback
```
mlissner: Can/should this callback be pulled into its own function?

Also, what happens if our upload fails? For example, if the RECAP server is down?

src/content_delegate.js
```diff
@@ -585,31 +588,193 @@ ContentDelegate.prototype.attachRecapLinkToEligibleDocs = function() {
...
+        }
+      );
     }
-  }, this));
+  } catch (err) {
+    console.error(err);
+  }
 };
+
+// Same as handlePDFView, but for zip files!
```
mlissner: I **think** you mean this:

```suggestion
// Same as handleSingleDocumentPageView, but for zip files
```

src/content_delegate.js
```diff
@@ -585,31 +588,193 @@ ContentDelegate.prototype.attachRecapLinkToEligibleDocs = function() {
...
+    '}',
+    'let items = document.getElementsByTagName("input");',
+    'for (i = 0; i < items.length; i++) {',
+    'input = items[i];',
+    'if (input.type === "button" && input.value === "Download Documents") {',
+    'input.removeAttribute("onclick");',
+    `input.addEventListener("click", () => window.postMessage({ id: ${JSON.stringify(url)}}))`,
+    '};',
+    '}'
+  ].join('');
```
mlissner: This is in lieu of the monkey patch approach for PDFs? Can you make a comment about how it's dangerous, what bad things could be done or happen, etc? (Assume I'm bad at this.)

src/content_delegate.js
```diff
@@ -144,65 +144,72 @@ ContentDelegate.prototype.findAndStorePacerDocIds = function() {
...
       }
     }
   }
-  chrome.storage.local.set(docsToCases, function(){
+
+  // identify current doc Number with simple key in chrome storage
+  const urlMatch = PACER.getDocumentIdFromUrl(document.URL);
+  if (!!urlMatch) {
+    saveItemToStorage({docId: urlMatch});
+  }
```
mlissner: I might have missed it, but does this get used somewhere? Sorry, I don't quite grok why you're doing this here.

src/content_delegate.js
mlissner: Narrator: He **was** pretty bad at it.

src/utils.js
```diff
@@ -140,8 +161,7 @@ $.ajaxSetup({
         // using this token. Unfortunately, there is no way to distribute
         // extensions that use hardcoded tokens except through begging and using
         // funny variable names. Do not abuse the RECAP service.
-        let asdf = '45c7946dd8400ad62662565cf79da3c081d9b0e5';
```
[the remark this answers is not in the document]
litewarp: ok

src/recap.js
litewarp: what would you prefer?

src/recap.js
litewarp: only thing that would be removed and possibly reused is the authorization header, but even if I were to abstract that into a global fetch option, that would also apply to calls made to the pacer sites, so I'm not sure we want that. For a post request, you need to pass in the POST flag and the payload to be uploaded.

src/content_delegate.js
litewarp: We can. I was trying to avoid callback hell by putting the callback directly in the call, but I'll abstract it beforehand.

litewarp: If the call falls, the catch in the try/catch will send the error to the console

src/content_delegate.js
litewarp: I mean the danger is pretty self evident, no? Injecting html means arbitrary code execution ... which is dangerous. I'll put a comment beforehand, but not sure how detailed you want me to be.

src/content_delegate.js
litewarp: Yes, the uploadPDF function pulls the docID from storage

src/recap.js
mlissner: Well, it's a dict with a couple entries, so something like storage_response might make sense? Or is there a clever way to do something like this in JS:

```
const bytes, docId = await getItemsFromStorage([nonce, 'docId'])
```

But the larger point was that the response from `getItemsFromStorage` isn't just a big blob of data, it's a dictionary with a few things in it. I was surprised to see that bytes had keys a few lines later.

src/content_delegate.js
mlissner: Thanks to both points.

src/content_delegate.js
mlissner: I mean, yes, injecting HTML/JS is bad, but there are no variables here, right, so I'm not sure I understand what an abuse/attack of this would look like.

src/content_delegate.js
mlissner: Looks like uploadPDF gets the pacer_doc_id passed as a parameter, and the uploadZip function gets it from storage. Why not pass it through for the zip function like we do for the PDF function. It looks like in content_delegate.js you should be able to just do `this.pacer_doc_id`? Maybe?

src/recap.js
mlissner: I see you've tweaked this. Seems better, thanks.

src/content_delegate.js
litewarp: There is some peculiarity in the flow where the this.pacer_doc_id isn’t set. I think the doc id is set two pages before the download function. So you’d have to set the docID and then persist it for further down the line. Don’t know how they did it before but it seems cleaner to put it into the global store and retrieve it from there.

src/content_delegate.js
litewarp: Yeah I’m agnostic about this. I copied this from the showPDFpage function. Feel free to take a crack as I won’t be back at my desk until later.

src/content_delegate.js
litewarp: Stylistic preference. Up to you.

src/content_delegate.js
litewarp: You’re giving a program the ability to write js that will execute immediately on the browser. That JS could access data on the page, potentially account credentials, or it could install or run malware, or hijack your browser to mine bitcoin ... hence why injecting JS as html is dangerously set

src/content_delegate.js
mlissner: OK, got it. Feels like there should be a better way to do this that doesn't use globally available persistence. I'm concerned about two things with this:

1. It's different than how we do PDFs, so it's two things to put in our heads.
2. It's not tied to the specific tab, so it could present race conditions even if a user is slow.

Imagine:

1. Go to a docket, choose a zip to download, stop at the receipt page.
2. New tab, go to a docket, choose a different zip to download, stop at the receipt page.
3. Go to the first tab and download it. 

That spells trouble, right? Could we use the tab `id` to tie the nonce to a specific tab: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/Tab

src/recap.js
```diff
@@ -122,65 +131,115 @@ function Recap() {
...
-          },
-          error: function(xhr, textStatus, errorThrown){
-            console.error(`RECAP: Ajax error uploading PDF. Status: ${textStatus}.` +
-                          `Error: ${errorThrown}`);
-          }
-        });
-
-      });
+          body: data,
+          headers: {'Authorization': `Token ${N87GC2}`}
```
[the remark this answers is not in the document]
litewarp: No. The browserSpecificFetch issue is because of a particular security restriction imposed by Firefox which prevents content scripts from calling fetch on a path ('/api/stuff') which we need to do to avoid hitting the "You are coming to Pacer from outside ... we will gouge you for monies". 

The recap functions are background functions that use the absolute recap URLs, so we can use the normal fetch mechanism.

PR #119, opened 2020-04-12
Add follow alert

litewarp:


src/content_delegate.js
```diff
@@ -242,50 +221,67 @@ ContentDelegate.prototype.handleDocketDisplayPage = async function () {
...
+  // If we don't have any pacerCaseId at this point, punt.
+  if (!pacerCaseId) { return; }
 
   const options = await getItemsFromStorage('options');
 
   if (options['recap_enabled']) {
-    let callback = $.proxy(function (ok) {
+    let callback = (ok) => {
       if (ok) {
+        const banner = document.querySelector('.recap-alert-banner');
```
mlissner, line 263: Back in the day, `recap_enabled` allowed you to completely turn off RECAP, but nowadays it only means that uploads are turned off. I think putting this behind the `recap_enabled` flag is a mistake then, because even if I have uploads off, I should be able to create an alert.

src/utils.js
```diff
@@ -217,3 +217,50 @@ function debug(level, varargs) {
     return console.log.apply(this, args);
   }
 }
+
+// inject a "follow this case on RECAP" button
+const recapButton = (court, pacerCaseId) => {
```
mlissner, line 222: Can we rename this as `createAlertButton` or something? recapButton vs. recapBanner has left me scratching my head a lot during this review.

src/utils.js
```diff
@@ -217,3 +217,50 @@ function debug(level, varargs) {
...
+  anchor.href = url.toString();
+  const img = document.createElement('img');
+  img.src = chrome.extension.getURL('assets/images/icon-16.png');
+  anchor.innerHTML = `${img.outerHTML} Create an alert for this case on RECAP`;
+  div.appendChild(anchor);
+
+  return div;
+};
+
+const recapBanner = (result) => {
```
mlissner, line 241: And maybe make this `createRecapDocketNotice` or something?

src/content_delegate.js
```diff
@@ -242,50 +221,67 @@ ContentDelegate.prototype.handleDocketDisplayPage = async function () {
...
 
   // if you've already uploaded the page, return
-  if (history.state && history.state.uploaded) {
-    return;
-  }
+  if (history.state && history.state.uploaded) { return; }
 
   // check if appellate
-  let isAppellate = PACER.isAppellateCourt(this.court);
+  // let isAppellate = PACER.isAppellateCourt(this.court);
```
mlissner, line 248: Why comment out? if we don't need it, let's remove it?

src/content_delegate.js
```diff
@@ -242,50 +221,67 @@ ContentDelegate.prototype.handleDocketDisplayPage = async function () {
   const radioDateInputs = [...document.getElementsByTagName('input')].filter(
     input => input.name === 'date_from' && input.type === 'radio'
   );
-  if (radioDateInputs.length > 1) {
-    return;
-  };
+  if (radioDateInputs.length > 1) { return; };
+
+  this.recap.getAvailabilityForDocket(
```
mlissner, line 226: The repetition between this and the other call to `getAvailablilityForDocket` bugs me, but more importantly, I don't think this needs to be behind an API call. I'm totally comfy with our users once in a while getting to the create alert page before the docket is truly ready. In the best case, the server will already have the item, but sometimes the server is really busy and processing can take time. I think we can just show the "create alert" button immediately and have them refresh on CL.

src/assets/css/style.css
```diff
@@ -136,6 +136,13 @@ footer #version {
 /*
    CSS for styling content on PACER pages generated by the content script.
 */
+.recap-alert-banner {
```
mlissner, line 139: This seems repetitive, but I suppose it's fine. Someday we can refactor our CSS, I guess, if it ever gets unwieldy.

PR #273, opened 2022-11-17
fix(docket report): Update logic to detect docket report 

ERosendo:
Fix [239](https://github.com/freelawproject/recap/issues/249)

PR addresses issues related to the docket report when it's accessed using the 'Reports' option in CM/ECF 1.6 and 1.7

Currently, the extension is not able to identify the page because the pacer_case_id is not part of the URL when the report loads.  The **case_id** is stored as the value of a hidden input with the id '**all_case_ids**' and changes when the user interacts with a text input used as a search bar for cases. 

The extension doesn't have the logic to listen to the changes being made to this value, so even when the **case_id** is inserted after the user specified a **case number**, the extension won't show the notice saying whether we have that docket or upload it. 

This PR adds a **listener** to the text input used as a search bar and a [Mutation observer](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) to the hidden input that stores the **case_id**. Both of these interfaces will provide the ability to watch for changes being made to the [DOM](https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model) so the extension can keep track of the interactions of the user and the changes in the **case_id** to respond accordingly.

This PR also fixes the tests in the **ContentDelegateSpec.js** file by mocking a couple of ajax calls using [jasmine-ajax](https://github.com/jasmine/jasmine-ajax), adding fake functions to replace the `getElementById` and `querySelector` methods and removing the function that was deleting the chrome object in the global context.

### DIFF
--- a/spec/ContentDelegateSpec.js
+++ b/spec/ContentDelegateSpec.js
@@ -623,6 +623,7 @@ describe('The ContentDelegate class', function () {
   describe('handleSingleDocumentPageCheck', function () {
     let form;
     beforeEach(function () {
+      clearDocumentBody()
       form = document.createElement('form');
       document.body.appendChild(form);
     });
--- a/spec/ContentDelegateSpec.js
+++ b/spec/ContentDelegateSpec.js
@@ -62,9 +62,6 @@ describe('The ContentDelegate class', function () {
       }
     }
   }
-  function removeChromeSpy() {
-    delete window.chrome;
-  }
 
   function clearDocumentBody() {
     document.body.innerHTML=''
@@ -83,7 +80,6 @@ describe('The ContentDelegate class', function () {
 
   afterEach(function () {
     jasmine.Ajax.uninstall();
-    removeChromeSpy();
     window.fetch = nativeFetch;
   });
 
@@ -280,7 +276,6 @@ describe('The ContentDelegate class', function () {
 
       afterEach(function () {
         table.remove();
-        delete window.chrome;
       });
 
       it('has no effect when recap_enabled option is false', function () {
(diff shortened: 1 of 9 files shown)


spec/ContentDelegateSpec.js
```diff
@@ -2,7 +2,7 @@
 describe('The ContentDelegate class', function () {
   // 'tabId' values
   const tabId = 1234;
-
+
```
mlissner, line 5: Nit, but try to be careful about this kind of tweak in a PR. I usually go line by line when doing my commits, and catch this kind of thing when I do.

src/pacer.js
```diff
@@ -87,6 +87,13 @@ let PACER = {
     return !!url.match(/\/(DktRpt|HistDocQry)\.pl\?\d+$/);
   },
 
+  // Returns the URL with the case id as a query parameter. This function makes
+  // sure every URL related to the Docket report has the same format
+  formatDocketQueryUrl: function(url, case_id){
+    return /[?&]/.test(url)  && /DktRpt.pl/.test(url)? url : `${url}?${case_id}`
+  },
```
mlissner, line 94: A comment here with a few examples of what's getting fixed would be helpful.

mlissner (changes requested):
A couple little things and I await your comments about the empty input we saw together, if that's affected by these changes somehow. 

I also added the other related issues to this so they'll close when we merge this (see the sidebar). And I moved the issues to your backlog, so they'll be in your done pile once merged.

Also, before merge:

 - [ ] Add task to the QA wiki

mlissner:
Thank you. Great stuff.

PR #274, opened 2022-11-18
feat(recap email): Add advertisement banners for RECAP Email

ERosendo:
Fixes [#309](https://github.com/freelawproject/recap/issues/309)

### DIFF
--- a/src/content_delegate.js
+++ b/src/content_delegate.js
@@ -182,12 +182,12 @@ ContentDelegate.prototype.findAndStorePacerDocIds = function () {
 
 // If this is a docket query page, add RECAP email advertisement.
 ContentDelegate.prototype.addRecapEmailAdvertisement = async function () {
-  if (!(PACER.isIQueryReportUrl(this.url) || PACER.isManageAccountPage(this.url))) { return; };
+  if (!(PACER.isBlankQueryReportUrl(this.url) || PACER.isManageAccountPage(this.url))) { return; };
   let form; 
 
   if (!PACER.hasFilingCookie(document.cookie)){ return; };
    
-  if (PACER.isIQueryReportUrl(this.url)){
+  if (PACER.isBlankQueryReportUrl(this.url)){
     form = document.querySelector('form');
     if (!document.querySelector('.recap-email-banner')){
       form.appendChild(recapEmailBanner());
--- a/src/content_delegate.js
+++ b/src/content_delegate.js
@@ -180,6 +180,27 @@ ContentDelegate.prototype.findAndStorePacerDocIds = function () {
   });
 };
 
+// If this is a docket query page, add RECAP email advertisement.
+ContentDelegate.prototype.addRecapEmailAdvertisement = async function () {
+  if (!(PACER.isIQueryReportUrl(this.url) || PACER.isManageAccountPage(this.url))) { return; };
+  let form; 
+
+  if (!PACER.hasFilingCooking(document.cookie)){ return; };
+   
+  if (PACER.isIQueryReportUrl(this.url)){
+    form = document.querySelector('form');
+    if (!document.querySelector('recap-email-banner')){
+      form.appendChild(recapEmailBanner());
+    }
+  }else{
+    form = document.querySelector('#popupForm');
+    if(!document.querySelector('recap-email-banner-full')){
+      form.after(recapEmailBanner('recap-email-banner-full'));
+    }
+  }
+  
+};
+
 // If this is a docket query page, ask RECAP whether it has the docket page.
 ContentDelegate.prototype.handleDocketQueryUrl = function () {
   if (!PACER.isDocketQueryUrl(this.url)) { return; };
--- a/src/content_delegate.js
+++ b/src/content_delegate.js
@@ -189,7 +189,7 @@ ContentDelegate.prototype.addRecapEmailAdvertisement = async function () {
    
   if (PACER.isIQueryReportUrl(this.url)){
     form = document.querySelector('form');
-    if (!document.querySelector('recap-email-banner')){
+    if (!document.querySelector('.recap-email-banner')){
       form.appendChild(recapEmailBanner());
     }
   }else{
--- a/src/content_delegate.js
+++ b/src/content_delegate.js
@@ -185,7 +185,7 @@ ContentDelegate.prototype.addRecapEmailAdvertisement = async function () {
   if (!(PACER.isIQueryReportUrl(this.url) || PACER.isManageAccountPage(this.url))) { return; };
   let form; 
 
-  if (!PACER.hasFilingCooking(document.cookie)){ return; };
+  if (!PACER.hasFilingCookie(document.cookie)){ return; };
    
   if (PACER.isIQueryReportUrl(this.url)){
     form = document.querySelector('form');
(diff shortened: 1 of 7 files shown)


src/pacer.js
```diff
@@ -400,6 +412,17 @@ let PACER = {
...
+  // Given document.cookie, returns true if the user has filing rights.
+  hasFilingCooking: function (cookieString){
+    let cookies = {};
+    cookieString.replace(/\s*([^=;]+)=([^;]*)/g, function (match, name, value) {
+      cookies[name.trim()] = value.trim();
+    });
+    let filingCookie = cookies['filing']
+    console.log(filingCookie)
+    return !!(filingCookie && filingCookie.match(/true/));
+  },
```
mlissner, line 424: There's gotta be a (much) better way. JavaScript sometimes surprises me, but are you sure this is the best way to do this??

src/pacer.js
```diff
@@ -197,6 +208,7 @@ let PACER = {
     let bigFile = document.getElementById('file_too_big')
     let buttonText = inputs.length ? inputs[inputs.length - 1].value.includes('Download') : false
     let mainContent = document.getElementById("cmecfMainContent");
+    if (!mainContent){ return false }
```
mlissner, line 211: When does this happen? What caused you to add this?

src/pacer.js
```diff
@@ -87,6 +87,16 @@ let PACER = {
     return !!url.match(/\/(DktRpt|HistDocQry)\.pl\?\d+$/);
   },
 
+  // Returns true if the URL is for the manage account page.
+  isManageAccountPage: function(url){
+    return /pacer./.test(url) && /manage/.test(url)
+  },
```
mlissner, line 93: Good to provide a comment with an example of the URL it matches.

src/utils.js
```diff
@@ -278,3 +288,20 @@ const recapBanner = (result) => {
...
+
+  const anchor = document.createElement('a');
+  anchor.title = 'Learn More'
+  anchor.target = '_blank';
+  anchor.href = `https://www.courtlistener.com/help/recap/email/`
+
+  const img = document.createElement('img');
+  img.src = chrome.extension.getURL('assets/images/icon-16.png');
+
+  anchor.innerHTML = `${img.outerHTML} Use @recap.email to automatically contribute all your cases to RECAP. Learn more`;
```
mlissner, line 304:
```suggestion
  anchor.innerHTML = `${img.outerHTML} Use @recap.email to automatically contribute all your cases to RECAP.`;
```

src/utils.js
```diff
@@ -278,3 +288,20 @@ const recapBanner = (result) => {
   div.appendChild(small);
   return div;
 };
+
+const recapEmailBanner = (css_class = 'recap-email-banner') => {
+  const div = document.createElement('div');
+  div.setAttribute('class', css_class);
+
+  const anchor = document.createElement('a');
+  anchor.title = 'Learn More'
```
mlissner, line 297: We can skip this. It's not a helpful pop up I don't think.

src/content_delegate.js
```diff
@@ -180,6 +180,27 @@ ContentDelegate.prototype.findAndStorePacerDocIds = function () {
...
+// If this is a docket query page, add RECAP email advertisement.
+ContentDelegate.prototype.addRecapEmailAdvertisement = async function () {
+  if (!(PACER.isIQueryReportUrl(this.url) || PACER.isManageAccountPage(this.url))) { return; };
+  let form; 
+
+  if (!PACER.hasFilingCooking(document.cookie)){ return; };
+   
+  if (PACER.isIQueryReportUrl(this.url)){
+    form = document.querySelector('form');
+    if (!document.querySelector('recap-email-banner')){
```
mlissner: Shouldn't this (and below) be:

```suggestion
    if (!document.querySelector('.recap-email-banner')){
```

src/pacer.js
```diff
@@ -87,6 +87,16 @@ let PACER = {
...
+
+  // Returns true if the URL is for the iQuery page.
+  isIQueryReportUrl: function(url){
```
mlissner, line 96: Per conversation on Slack, let's rename this as `isBlankQueryReportUrl`, and then add a comment that explains the difference between the two types of URLs that this matches, and why.

src/utils.js
```diff
@@ -184,6 +184,16 @@ const saveCaseIdinTabStorage = async (object, case_id) => {
   })
 }
 
+// Save a cookie in document.cookie to let the extension know that the user has filing rights
+const setFilingState = () => {
+  document.cookie = "filing=true;path=/;domain=.uscourts.gov";
```
mlissner, line 189: I think I would suggest `isFilingAccount` as your cookie name. Much more clear.

src/content_delegate.js
```diff
@@ -180,6 +180,27 @@ ContentDelegate.prototype.findAndStorePacerDocIds = function () {
...
+  if (!PACER.hasFilingCooking(document.cookie)){ return; };
+   
+  if (PACER.isIQueryReportUrl(this.url)){
+    form = document.querySelector('form');
+    if (!document.querySelector('.recap-email-banner')){
+      form.appendChild(recapEmailBanner());
+    }
+  }else{
+    form = document.querySelector('#popupForm');
+    if(!document.querySelector('recap-email-banner-full')){
```
mlissner, line 197: You still need to tweak this class, right?

mlissner (changes requested):
A few little things.

mlissner:
Was it too annoying to add this to the settings page too? I'm guessing you couldn't access the cookie value? I actually thought you'd set the cookie to the background page's context, not the content page's. I'm not sure if that's easier or not though (and anyway, it seems like it's working as is).

src/pacer.js
ERosendo: You're right! I think We can use the approach shown in the [MDN Web Docs](https://developer.mozilla.org/en-US/). Here's the piece of code used to get the value of a cookie named test2:

```javascript
const cookieValue = document.cookie
  .split('; ')
  .find((row) => row.startsWith('test2='))
  ?.split('=')[1];
```

The whole example is [here](https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie#example_2_get_a_sample_cookie_named_test2).

mlissner: Yeah, that's better, but man, JS just really sucks sometimes!

src/content_delegate.js
ERosendo: You're right!

changed. Thanks

src/pacer.js
ERosendo: The `cmecfMainContent` is the parent `div` that encloses almost all the `forms` in **CM/ECF**, but that `div `doesn't exist in the PACER website so the **extension** always gets `null ` while the user is using one of the pages (like Manage my account) in websites like https://pacer.psc.uscourts.gov/.

if the value of mainContent is null then the extension will find an Uncaught typeError in line 227 of that file when it tries to get the `lastChild` of the div. I added the validation before it reaches the possible typeError,

src/pacer.js
```diff
@@ -197,6 +223,7 @@ let PACER = {
     let bigFile = document.getElementById('file_too_big')
     let buttonText = inputs.length ? inputs[inputs.length - 1].value.includes('Download') : false
     let mainContent = document.getElementById("cmecfMainContent");
+    if (!mainContent){ return false }
```
mlissner, line 226:
```suggestion
    // End this function early if we're on a management PACER page
    if (!mainContent){ return false }
```

mlissner (approved):
Looks good, thanks.

PR #275, opened 2022-11-21
fix(recap popup): Remove recap popup and update anchors href

ERosendo:
This PR solves multiple issues related to the links inserted by the RECAP extension:

- Avoid creating links that redirect to storage.courtlistener.com:
  The links inserted by the extension go to [www.courtlistener.com](http://www.courtlistener.com/), which redirects to `storage.courtlistener.com`. This PR will change the links used in the href of the anchor tags to avoid redirection.  
- Remove the option to use a popup before opening RECAP documents
  The current implementation of the extension has a check box on the options page that says, "Confirm before opening RECAP documents". If it's checked( which it is by default) it pops up a modal whenever you click an [R] icon to download a document. This PR fixes [#216](https://github.com/freelawproject/recap/issues/216) by removing all the logic related to this popup and updating the test in the ContentDelegateSpec file.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,7 +8,7 @@ Features:
  - None yet
 
 Changes:
- - None yet
+ - Remove option to "Confirm before opening RECAP documents" ([#216](https://github.com/freelawproject/recap/issues/216))
 
 Fixes: 
  - None yet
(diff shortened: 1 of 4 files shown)


mlissner:
Gotta love a PR that just removes code!

PR #277, opened 2022-11-23
feat(iquery page): Send iquery summary pages

ERosendo:


src/content_delegate.js
```diff
@@ -364,6 +364,43 @@ ContentDelegate.prototype.handleAttachmentMenuPage = function () {
...
+          'iQuery page uploaded to the public RECAP Archive.',
+          function () {
+          }
+        );
+      }
+    }, this);
+
+    this.recap.uploadIQueryPage(this.court, this.pacer_case_id,
+      document.documentElement.innerHTML, callback)
+  }
```
mlissner, line 401: We should have an `else` here like on line 361, right?

src/pacer.js
```diff
@@ -288,6 +300,24 @@ let PACER = {
...
+    let anchors = tableContainer.getElementsByTagName("a");
+    // End this function early if there are no anchor tags
+    if (!anchors.length){ return null; }
+    let lastAnchor = anchors[anchors.length-1]
+    // the next secuence of instruccions will match the digits at the end of the href. 
+    // The href attribute of the anchor tags in the table has the following format:
+    //
+    //    https://ecf.mied.uscourts.gov/cgi-bin/qryDocument.pl?360406
+    //
+    // We can grab the pacer case ID from this url if we get all the digits after the ?.
```
mlissner, line 315: Check your spelling here, please. I recommend turning on spellchecker in English if you don't have it on in your editor already.

mlissner (changes requested):
A few little things, please.

src/content_delegate.js
ERosendo: You're right!

changed. Thanks

PR #285, opened 2022-12-02
fix(links from appellate): Makes isDocketQueryUrl more generic

ERosendo:
This PR fixes https://github.com/freelawproject/recap/issues/222

The current implementation of the isDocketQueryUrl function only matches URLs that have a query string with only digits. That was fine, but the link from the appellate pacer to the original case has a different query string that includes the caseNumber parameter, so doesn't match properly in that case. 

The new regex expression will match both cases. Here's a gif showing that the extension will be able to insert banners in the `DktRpt.pl` after following the link in the appellate pacer.

recap-originating-case-number

### DIFF
--- a/src/pacer.js
+++ b/src/pacer.js
@@ -84,10 +84,18 @@ let PACER = {
       return match[0].slice(1);
     }
   },
+  
   // Returns true if the URL is for docket query page.
   isDocketQueryUrl: function (url) {
-    // The part after the "?" is all digits.
-    return !!url.match(/\/(DktRpt|HistDocQry)\.pl\?\d+$/);
+    // This regex expression will match URLs that has a query string with only
+    // digits or has a parameter named caseNumber followed by word characters, 
+    // hyphens or colons. This function will return true for the following URLs:
+    // 
+    //  https://ecf.mied.uscourts.gov/cgi-bin/DktRpt.pl?365816
+    //  
+    //  https://ecf.moed.uscourts.gov/cgi-bin/DktRpt.pl?caseNumber=4:19-cr-00820-SRC-1
+    //
+    return !!url.match(/\/(DktRpt|HistDocQry)\.pl\?(\d+|caseNumber=[\w:-]+)$/);
   },
 
   // Returns true if the URL is for the manage account page.
--- a/src/pacer.js
+++ b/src/pacer.js
@@ -87,16 +87,23 @@ let PACER = {
   
   // Returns true if the URL is for docket query page.
   isDocketQueryUrl: function (url) {
-    // This regex expression will match URLs that has a query string with only
-    // digits or has one or multiple parameters separated by the ampersand ("&").
+    // There are two pages that use the DktRpt.pl url that we want to match:
+    //  - The docket query page: /DktRpt.pl?115206
+    //  - The docket report page: /DktRpt.pl?caseNumber=1:17-cv-10577&caseId=0
+    // 
+    // And one page that uses DktRpt.pl that we don't want to match:
+    //  - The docket page itself: /DktRpt.pl?591030040473392-L_1_0-1
+    // 
+    // Thus, this regex matches URLs that have a query string with only digits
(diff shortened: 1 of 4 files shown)


mlissner (commented):
Seems like this is still pretty fragile. Why not do a regex that's much more lenient?

Second, we should have new tests for this, right?

src/pacer.js
```diff
@@ -84,10 +84,21 @@ let PACER = {
...
+    //
+    // This function will return true for the following URLs:
+    // 
+    //  https://ecf.mied.uscourts.gov/cgi-bin/DktRpt.pl?365816
+    //  
+    //  https://ecf.moed.uscourts.gov/cgi-bin/DktRpt.pl?caseNumber=4:19-cr-00820-SRC-1
+    //
+    //  https://ecf.mad.uscourts.gov/cgi-bin/DktRpt.pl?caseNumber=1:17-cv-10577&caseId=0
+    //
+    return !!url.match(/\/(DktRpt|HistDocQry)\.pl\?(\d+|(&?[\w]+=[^&=\n]+)+)$/);
```
mlissner, line 101: I'm still wondering why we shouldn't just be more lenient here. Why not something as simple as:

```suggestion
    return !!url.match(/\/(DktRpt|HistDocQry)\.pl\?.*/);
```

Would that over-include something we're trying to exclude?

ERosendo: The extension uses that method to identify the **Docket Query page** and the **Docket Report Page**. The URL for the **Docket Query page** is the one that has only digits (e.g. `/DktRpt.pl?115206`) and the **Docket Report page** could have a query string with one or more parameters if the user clicks the link available in appellate PACER ( e.g. `/DktRpt.pl?caseNumber=2:16-cv-01129-RFB-DJA` or `/DktRpt.pl?caseNumber=1:17-cv-10577&caseId=0` according to https://github.com/freelawproject/recap/issues/222). Here are the screenshots for these two pages:


**Docket Query page**  |  **Docket Report Page**
:-------------------------:|:-------------------------:
image |  image

There's another page with an URL ending in `DktRpt.pl` but has a different format in its query string, so it is excluded by this function, The **docket display page** has a URL  like this `/DktRpt.pl?591030040473392-L_1_0-1`. Here's the screenshot of that page:

image
 

Using a regex that matches any character after the start of the query string will return `true` for the **docket display page**.

mlissner: Got it, thank you. Let's just clarify this in the comment, then and we should be good.

mlissner: I'll suggest some language. :)

mlissner, line 101:
```suggestion
    // There are two pages that use the DktRpt.pl url that we want to match:
    //  - The docket query page: /DktRpt.pl?115206
    //  - The docket report page: /DktRpt.pl?caseNumber=1:17-cv-10577&caseId=0
    // 
    // And one page that uses DktRpt.pl that we don't want to match:
    //  - The docket page itself: /DktRpt.pl?591030040473392-L_1_0-1
    // 
    // Thus, this regex matches URLs that have a query string with only digits
    // or that has one or multiple parameters separated by the ampersand ("&").
    //
    // This function will return true for the following URLs:
    //  - DktRpt.pl?365816
    //  - DktRpt.pl?caseNumber=4:19-cr-00820-SRC-1
    //  - DktRpt.pl?caseNumber=1:17-cv-10577&caseId=0
    //
    // But false for:
    //  - /DktRpt.pl?591030040473392-L_1_0-1
    //
    return !!url.match(/\/(DktRpt|HistDocQry)\.pl\?(\d+|(&?[\w]+=[^&=\n]+)+)$/);
```

mlissner: @ERosendo let me know if you like that change. If so, please accept it and feel free to merge this PR. Thank you for helping me understand. It's tricky!

ERosendo: this comment is a lot nicer. Thanks for your help!

mlissner: It's the only way I can keep up with y'all.

PR #286, opened 2022-12-08
feat(appellate): Add support for the Case Summary Page from Appellate PACER

ERosendo:
This PR adds support for the Case Summary Page from Appellate PACER to the RECAP extension. 

Here's a picture of the RECAP actions button inserted on the page:

image

- The old logic that was created in previous attempts to add support for appellate pacer will be moved into a new utils file inside a folder named appellate 
- A new class named `AppellateDelegate` will be added and It will handle the identification of the different pages from Appellate and the DOM manipulation.
- This new class will have a handler method for each page that the extension is going to support. This PR will include two handlers, one for the **Case Summary Page** and another one for the **Case Selection Page** ( We need this handler now because the extension is able to find the `pacer_case_id` on this page and the user visits this page before reaching the Case Summary, so we can save the `case_id` to the chrome local storage and use it later. This handler will extend its functionality in a future PR)
- The new class will have a dispatcher function to choose the proper handler in the class.
- A new file with test related to the class and its utils will be added.

The changes included in this PR will fix https://github.com/freelawproject/recap/issues/314


src/appellate/appellate.js
```diff
@@ -0,0 +1,162 @@
...
+  // if you've already uploaded the page, return
+  if (history.state && history.state.uploaded) { return; }
+
+  const options = await getItemsFromStorage('options');
+
+  if (options['recap_enabled']) {
+
+    let callback = (ok) => {
+      if (ok) {
+        addAlertButtonInRecapAction(this.court, pacer_case_id)
```
mlissner, line 131: Does this belong here? Didn't we already do this 15 lines up? Why do we do it again?

src/content.js
```diff
@@ -15,13 +15,25 @@ let links = document.body.getElementsByTagName('a');
...
+  const { tabId } = msg;
+
+  if (PACER.isAppellateCourt(court)) {
+
+    const AppDelegate = new AppellateDelegate(tabId, court, url, links);
+
+    if (!PACER.hasPacerCookie(document.cookie)) {
+      console.info(`RECAP: Taking no actions because not logged in: ${url}`);
+      return;
+    }
```
mlissner, line 29: Can we move this outside of this if statement as well as out of the else statement below?

src/content.js
```diff
@@ -15,13 +15,25 @@ let links = document.body.getElementsByTagName('a');
...
+
+    AppDelegate.dispatchPageHandler();
+  } else {
+
     let content_delegate = new ContentDelegate(tabId,
       url, path, court, pacer_case_id, pacer_doc_id, links);
```
mlissner, line 35: For consistency, shouldn't we refactor this to look like the appellate example, instead having all this logic in content.js, we run a `dispatchPageHandler` function on ContentDelegate? 

Also shouldn't ContentDelegate be renamed to be district court specific (like AppellateDelegate is)?

mlissner (changes requested):
This looks pretty good, but you'll see a bunch of comments. I also noticed that a lot of your code lacks needed semi-colons. I'm surprised your editor isn't yelling at you, and I suspect this means you're not running prettier. Could you do that, please and enable it going forward?

ERosendo:
@mlissner I've just realized there's a .prettierignore and It's ignoring all the .js files

src/appellate/appellate.js
```diff
@@ -0,0 +1,162 @@
...
+AppellateDelegate.prototype.dispatchPageHandler = function () {
+  if (APPELLATE.isCaseSelectionPage(this.url)) {
+    this.handleCaseSelectionPage()
+  } else {
+    let queryParameters = APPELLATE.getQueryParameters(this.url);    
+    if (!(queryParameters instanceof URLSearchParams)) {
+      return;
+    }
+
+    this.queryParameters = APPELLATE.getQueryParameters(this.url);
```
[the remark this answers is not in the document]
ERosendo: You're right. Thanks.

Changed.

src/appellate/appellate.js
```diff
@@ -0,0 +1,162 @@
...
+
+
+
+
+
+
+
+
+
+
```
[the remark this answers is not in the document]
ERosendo: You're right. The code review before committing to git makes a big difference in the quality of the work.
 
Thanks for your suggestion. I'll be more careful next time. 

I will check Prettier configuration to avoid these blank lines.

mlissner:
This looks good. Still a couple unresolved comments, please.

ERosendo:
@mlissner All of the conversations should be resolved by now.

Thanks.

ERosendo (commented):
Here are some comments about your suggestions.

src/content.js
mlissner: OK, great.

mlissner (approved):

PR #287, opened 2022-12-09
fix(banners): Add logic to avoid duplicating RECAP banners

ERosendo:
This PR fixes https://github.com/freelawproject/recap/issues/318. We use the solution suggested in the issue.


src/content_delegate.js
```diff
@@ -220,6 +220,7 @@ ContentDelegate.prototype.handleDocketQueryUrl = function () {
         );
       } else {
         if (result.results) {
+          PACER.removeBanners()
```
mlissner, line 223: Curious...no semi-colon here?

mlissner (commented):
Looks fine except for the very small nitpick. If you want to fix that and merge, that seems good to me.

src/content_delegate.js
ERosendo: I created the commit before We fixed the issue with Prettier. After I merged the [main branch](https://github.com/freelawproject/recap-chrome/tree/main) with this one, the code formatter made a lot of changes to this file, so I created a [.git-blame-ignore-revs](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view) file to avoid messing with the blame view.

Thanks for mentioning that file during the meeting.

PR #289, opened 2022-12-13
feat(attachment page): Add support for attachment page from Appellate PACER

ERosendo:
This PR adds support for the attachment page from Appellate PACER to the RECAP extension.

- This PR adds a new parameter called **upload_type** to the **uploadAttachmentMenu** method. The current implementation of this method only uploads attachment pages from district PACER.  This new parameter will allow us to use the same function to upload attachment pages from appellate PACER.

-  A new method to get the `Servlet` parameter will be added. This new method will use hidden `inputs` on the page to get it.

- This PR will refactor the  **dispatchHandler** method. The new method to get the Servlet parameter will allow us to easily identify the **case selection page**.

- The current implementation of the `AppellateDelegate` class is not able to upload the **full docket page**, this class is not able to identify it using the page's URL but, this PR will fix that issue using the new method that gets the `Servlet` parameter from inputs on the page.

- Two new functions to get the `caseId` will be added. One of these functions uses the query string of the URL while the other one tries to get this id from inputs on the page.

- This PR will add a helper function that creates an iframe element on the **case summary** and **full docket page**. We will use the iframe to change the default behavior of a hidden `form` on these pages. This hidden form is submitted when an `anchor` element related to a document is clicked and its **target attribute** is set to `_blank`, so a tab is open every time we click one of those links. Changing the target attribute of the `form` will allow us to keep using the same tab ( using the same tab is preferred because the RECAP extension uses the `tab id` as a key to save data to the local storage. If a document is open in a different tab, the extension won't be able to get the data that was previously saved in the local storage).

- The logic used to filter links related to documents on the page will be updated to add the creation of a listener that avoids opening new tabs when these links are clicked.

The changes included in this PR will fix https://github.com/freelawproject/recap/issues/315

### DIFF
--- a/src/appellate/appellate.js
+++ b/src/appellate/appellate.js
@@ -164,7 +164,7 @@ AppellateDelegate.prototype.handleDocketDisplayPage = async function () {
 AppellateDelegate.prototype.handleAttachmentPage = async function () {
   this.pacer_case_id = await APPELLATE.getCaseId(this.tabId, this.queryParameters, this.docId);
 
-  if (this.pacer_case_id) {
+  if (!this.pacer_case_id) {
     return;
   }
 
--- a/src/appellate/appellate.js
+++ b/src/appellate/appellate.js
@@ -43,11 +43,11 @@ AppellateDelegate.prototype.handleCaseSelectionPage = async function () {
 AppellateDelegate.prototype.attachRecapLinksToEligibleDocs = async function () {
   // When you click a document link, it runs JS that submits this form.
   // Here, we override the target attribute of the form. If you don't do
-  // this, the form opens a new tab (target="_blank" by default), and 
-  // we would be unable to link that tab back to the metadata we 
-  // captured here. Thus, by overriding this form, we are able to 
+  // this, the form opens a new tab (target="_blank" by default), and
+  // we would be unable to link that tab back to the metadata we
+  // captured here. Thus, by overriding this form, we are able to
   // maintain the context we need to upload docs to the archive.
-  
+
   let form = document.getElementsByName('doDocPostURLForm');
   if (form.length) {
     form[0].setAttribute('target', '_self');
--- a/src/appellate/appellate.js
+++ b/src/appellate/appellate.js
@@ -35,13 +35,12 @@ AppellateDelegate.prototype.dispatchPageHandler = function () {
 
 AppellateDelegate.prototype.handleCaseSelectionPage = async function () {
   // retrieve pacer_case_id from the link related to the Case Query
-  let pacer_case_id = APPELLATE.getCaseIdFromCaseSelection();
-  await saveCaseIdinTabStorage({ tabId: this.tabId }, pacer_case_id);
+  this.pacer_case_id = APPELLATE.getCaseIdFromCaseSelection();
+  await saveCaseIdinTabStorage({ tabId: this.tabId }, this.pacer_case_id);
 };
 
 // check every link in the document to see if RECAP has it
 AppellateDelegate.prototype.attachRecapLinksToEligibleDocs = async function () {
- 
   let form = document.getElementsByName('doDocPostURLForm');
   if (form.length) {
     form[0].setAttribute('target', '_self');
@@ -49,6 +48,14 @@ AppellateDelegate.prototype.attachRecapLinksToEligibleDocs = async function () {
 
   // filter the links for the documents available on the page
   let { links, docsToCases } = APPELLATE.findDocLinksFromAnchors(this.links);
+
+  this.pacer_case_id = this.pacer_case_id
+    ? this.pacer_case_id
+    : await APPELLATE.getCaseId(this.tabId, this.queryParameters, this.docId);
+
+  if (this.pacer_case_id && this.docId) {
+    docsToCases[this.docId] = this.pacer_case_id;
+  }
 
   updateTabStorage({
     [this.tabId]: {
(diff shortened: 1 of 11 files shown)


src/appellate/appellate.js
```diff
@@ -86,7 +98,7 @@ AppellateDelegate.prototype.attachRecapLinksToEligibleDocs = async function () {
 
 AppellateDelegate.prototype.handleDocketDisplayPage = async function () {
   // retrieve pacer_case_id from URL query parameters
```
mlissner, line 100: This comment is out of date. I think you can probably just remove it, right? The code is really clear about what it's doing.

src/appellate/appellate.js
```diff
@@ -133,9 +145,49 @@ AppellateDelegate.prototype.handleDocketDisplayPage = async function () {
...
+  let pacer_case_id = APPELLATE.getCaseIdFromSearchQuery(this.queryParameters) || APPELLATE.getCaseIdFromInputs();
+
+  // if the last step didn't find the caseId in the query parameter, It will check the storage
+  if (!pacer_case_id) {
+    const tabStorage = await getItemsFromStorage(this.tabId);
+    if (!tabStorage && !tabStorage.caseId) {
+      return;
+    }
+    pacer_case_id = tabStorage.caseId;
+  }
```
mlissner, line 167: Seeing this in a lot of places makes me think we should just have one utility function for this. Something like:

```
getCaseID: (queryParameters) => {
  // Do the search query part
  ...
  // Then do the from inputs part
  ...
  // Then do the from storage part
  ...
}
```

src/appellate/appellate.js
```diff
@@ -36,6 +41,13 @@ AppellateDelegate.prototype.handleCaseSelectionPage = async function () {
 // check every link in the document to see if RECAP has it
 AppellateDelegate.prototype.attachRecapLinksToEligibleDocs = async function () {
   // filter the links for the documents available on the page
+  APPELLATE.createDummyIframe('dummyframe');
+
+  let form = document.getElementsByName('doDocPostURLForm');
+  if (form.length) {
+    form[0].setAttribute('target', 'dummyframe');
+  }
```
mlissner, line 49: Per discussion today on Slack, it seems like this and related code can be simplified by just removing the `target` attribute of the `doDocPostURLForm` form.

mlissner (changes requested):
OK, good stuff. The one big thing is the simplification of the form handling we discussed on Slack. The rest looks solid.

mlissner:
Also, I'm a bit concerned that we're changing the functionality of PACER to not open new tabs. We generally try pretty hard to avoid that, but I think the tradeoff is worth it, and I think the default functionality is weird anyway (though others may disagree). I suspect that a fair number of people are used to it opening new tabs and will CTRL+Click links in order to maintain the old functionality. When they do, I guess RECAP won't work, which is too bad, but if PACER keeps working, I think that's OK.

src/appellate/appellate.js
ERosendo: You're right. Thanks

Changed

src/appellate/appellate.js
ERosendo: You're right! A utility function will simplify things.

Thanks

ERosendo:
I've applied the requested changes, including:

- Creating a utility function called `getCaseId` that uses different approaches to retrieve the `pacer_case_id`
- Remove the dummy iframe and set the target attribute of the form to `_self`. I thought I could remove the listener that the extension is creating to submit the form but after some testing, I noticed that using the default listener the PACER uses for the links removes the servlet parameter in the URL for the attachment page, so I kept the new listener because We're using the servlet parameter to get the docId for the attachment page.
- Creating an object that uses the docId as a key and the pacerCaseId as values and saves it to the tab storage, so we can map a docId to a pacerCaseId when the extension is handling pages related to documents and they don't have the caseId available on the HTML or in the URL ( like the attachment page)

src/appellate/appellate.js
```diff
@@ -124,18 +143,53 @@ AppellateDelegate.prototype.handleDocketDisplayPage = async function () {
...
-    this.attachRecapLinksToEligibleDocs();
   } else {
     console.info(`RECAP: Not uploading docket. RECAP is disabled.`);
   }
 };
+
+AppellateDelegate.prototype.handleAttachmentPage = async function () {
+  this.pacer_case_id = await APPELLATE.getCaseId(this.tabId, this.queryParameters, this.docId);
+
+  if (this.pacer_case_id) {
```
mlissner, line 167: Shouldn't this be:

```suggestion
  if (!this.pacer_case_id) {
```

src/appellate/utils.js
```diff
@@ -48,12 +92,53 @@ let APPELLATE = {
...
+      if (!PACER.isDocumentUrl(a.href)) return;
+
+      // this regex will match the doc_id and case_id passed as an argument in the
+      // onclick event of each anchor element related to a document
+      let doDocPost = /^return doDocPostURL\('([^']*)','([^']*)'\);/.exec(a.getAttribute('onclick'));
+      let params = {};
+      if (doDocPost) {
+        [, params.doc_id, params.case_id] = doDocPost;
+        docsToCases[params.doc_id] = params.case_id;
+      }
```
mlissner, line 107: This can easily go in a helper function, where it'll be a lot easier to test.

src/appellate/utils.js
```diff
@@ -48,12 +92,53 @@ let APPELLATE = {
...
+
+      // add a new listener that allows the anchors to target the active tab
+      clonedNode.addEventListener('click', (event) => {
+        let form = document.getElementsByName('doDocPostURLForm')[0];
+        if (form) {
+          form.dls_id.value = params.doc_id;
+          form.caseId.value = params.case_id;
+          form.submit();
+        }
+      });
```
mlissner, line 124: How much of this is still needed if we update the form's `target` attribute so it doesn't open new tabs? I thought this would be unneeded if we did that, and we could just let the normal JS run without creating our own listeners.

mlissner (changes requested):
A couple more comments, but we're close.

src/appellate/appellate.js
ERosendo: Yes! You're right. That line should check if the `pacer_case_id` is empty. I will change this. 

Thanks

src/appellate/appellate.js
```diff
@@ -29,14 +35,33 @@ AppellateDelegate.prototype.dispatchPageHandler = function () {
...
+  this.pacer_case_id = APPELLATE.getCaseIdFromCaseSelection();
+  await saveCaseIdinTabStorage({ tabId: this.tabId }, this.pacer_case_id);
 };
 
 // check every link in the document to see if RECAP has it
 AppellateDelegate.prototype.attachRecapLinksToEligibleDocs = async function () {
+  let form = document.getElementsByName('doDocPostURLForm');
+  if (form.length) {
+    form[0].setAttribute('target', '_self');
+  }
```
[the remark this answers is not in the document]
ERosendo: I agree!! This part is really important. The comment you suggested explains it really well. I will add this. 

thanks

src/appellate/utils.js
ERosendo: You're right! I will create the helper function and add new tests,

src/appellate/utils.js
ERosendo: After changing the form's target attribute, I thought I could remove our listener too but then I tested a new implementation of the extension that used the default JS and noticed that the URL for the attachment page didn't include the **servlet** parameter. 

The problem with the URL that doesn't have the servlet parameter is that we don't have a reliable way to get the **case_id** from the attachment page. We can use the tab storage to retrieve the metadata we've captured and use the **case_id** from the previous page( **General Docket** or **Case Summary**) but I'm sure if an attachment page could have a different **case_id** than the one found in the previous pages ( could the attachment page  belong to another docket but is listed on the one we're reviewing?)

I didn't remove the new listener because I think the **servlet** parameter is important for the **attachment page**. this parameter includes the **doc_id** for this page. We can extract it with a regex expression and then use the mapping ( we create this object parsing the `onclick` event from the document anchor elements and holds the **doc_id** and the **case_id** as a key-value pair) to get the **case_id**. I think this approach is better because we're capturing the **case_id** for each document so we could have different ids inside the mapping.

mlissner: We concluded on Slack that documents can belong to multiple cases and that some cases just reference docs in other cases in their text. So we can't assume that attachment pages are from the same case as the docket that leads to them.

mlissner (approved):

PR #291, opened 2022-12-16
Free.law 105 fundraising mode

mlissner:


mlissner:
This should get merged after the appellate doc upload is in tomorrow.

ERosendo (commented):
I tested these changes in my local development environment and they worked as expected. If you want to merge, that seems good to me.

PR #294, opened 2022-12-20
feat(caseQueryAdvanced): Add support to identify this page from PACER

ERosendo:
This PR adds support to identify and upload the caseQueryAdvanced page from Appellate and District PACER. 

- This PR will add new upload types to the `upload_types `array in the recap class.  These new types are `CASE_QUERY_RESULT_PAGE` and  `APPELLATE_CASE_QUERY_RESULT_PAGE`.

- A new parameter called `upload_type` will be added to the uploadIQueryPage method. The current implementation of this method only uploads **query pages** from district PACER.  This new parameter will allow us to use the same function to upload **CaseQueryAdvanced** from district PACER and **caseQuery** pages from appellate PACER.

- This PR will add new helper functions to identify the **CaseQueryAdvanced** page. These helpers use HTML elements to detect this page because relying on the URL is not enough. We noticed that this page showed the transaction receipt table at the bottom of the page and also had the title "Select a case" at the top, so we created a utility function to check if the body of the page contains the  transaction receipt table ( `hasTransactionReceipt`) and another to check the title of the page and the description in the transaction receipt table( `isCaseQueryAdvance`)

- This PR will use the new helper functions to identify the Case Query Advance page and avoid uploading this page using the wrong upload_type. 

The changes included in this PR will fix https://github.com/freelawproject/recap/issues/321 and resolve https://github.com/freelawproject/courtlistener/issues/2419.

### Additional comments 

- When a user tries to search a case using only the first/middle/last name in the iquery form ([iquery.pl](https://ecf.mied.uscourts.gov/cgi-bin/iquery.pl)), they might get redirected to a list of matching people(**"Select a Person"** page) before they reach the list of cases related to that name. This PR will avoid uploading this page because the `pacer_case_id` cannot be extracted from the HTML elements. Here's a screenshot of this page:

image

- Some screenshots of the pages we're identifying and uploading are shown below: 

**Case Query Advance from District PACER**
image

**Case Query Advance from Bankruptcy PACER**
image

**Case Query Advance from Appellate PACER**
image

### DIFF
--- a/src/appellate/appellate.js
+++ b/src/appellate/appellate.js
@@ -34,14 +34,27 @@ AppellateDelegate.prototype.dispatchPageHandler = function () {
 };
 
 AppellateDelegate.prototype.handleCaseSelectionPage = async function () {
-  // retrieve pacer_case_id from the link related to the Case Query
-  if (document.getElementsByName('csnum1')[0].value || document.getElementsByName('csnum2')[0].value){
+ 
+  if (document.querySelectorAll('input:not([type=hidden])').length) {
+    // When the users go back to the Case Selection Page from the Docket Report using the back button
+    // Appellate PACER loads the Case Search Page instead but in the HTML body has the servlet hidden input
+    // and shows 'CaseSelectionTable.jsp' as its value.
+    //
+    // This check avoids sending pages like the one previously described to the API.
+    return;
+  }
+
+  if (APPELLATE.caseSelectionPageHasOneRow()) {
+    // Retrieve pacer_case_id from the Case Query link
     this.pacer_case_id = APPELLATE.getCaseIdFromCaseSelection();
     await saveCaseIdinTabStorage({ tabId: this.tabId }, this.pacer_case_id);
+  } else {
+    // Add the pacer_case_id to each docket link to use it in the docket report
+    APPELLATE.addCaseIdToDocketSummaryLink();
   }
-  
+
   const options = await getItemsFromStorage('options');
-  
+
   if (!options['recap_enabled']) {
     console.info('RECAP: Not uploading case selection page. RECAP is disabled.');
     return;
@@ -56,7 +69,7 @@ AppellateDelegate.prototype.handleCaseSelectionPage = async function () {
 
   this.recap.uploadIQueryPage(
     this.court,
-    null,
+    this.pacer_case_id,
     document.documentElement.innerHTML,
     'APPELLATE_CASE_QUERY_RESULT_PAGE',
     callback
--- a/src/appellate/appellate.js
+++ b/src/appellate/appellate.js
@@ -161,6 +161,8 @@ AppellateDelegate.prototype.handleDocketDisplayPage = async function () {
     return;
   }
 
+  await saveCaseIdinTabStorage({ tabId: this.tabId }, this.pacer_case_id);
+
   // Query the first table with case data and insert the RECAP actions button
   let table = document.querySelectorAll('table')[3];
   let button = recapActionsButton(this.court, this.pacer_case_id, false);
--- a/src/appellate/appellate.js
+++ b/src/appellate/appellate.js
@@ -35,8 +35,33 @@ AppellateDelegate.prototype.dispatchPageHandler = function () {
 
 AppellateDelegate.prototype.handleCaseSelectionPage = async function () {
   // retrieve pacer_case_id from the link related to the Case Query
-  this.pacer_case_id = APPELLATE.getCaseIdFromCaseSelection();
-  await saveCaseIdinTabStorage({ tabId: this.tabId }, this.pacer_case_id);
+  if (document.getElementsByName('csnum1')[0].value || document.getElementsByName('csnum2')[0].value){
+    this.pacer_case_id = APPELLATE.getCaseIdFromCaseSelection();
+    await saveCaseIdinTabStorage({ tabId: this.tabId }, this.pacer_case_id);
+  }
+  
+  const options = await getItemsFromStorage('options');
+  
+  if (!options['recap_enabled']) {
+    console.info('RECAP: Not uploading case selection page. RECAP is disabled.');
+    return;
+  }
+
+  let callback = (ok) => {
+    if (ok) {
+      history.replaceState({ uploaded: true }, '');
+      this.notifier.showUpload('Case selection page uploaded to the public RECAP Archive.', function () {});
+    }
+  };
+
+  this.recap.uploadIQueryPage(
+    this.court,
+    null,
+    document.documentElement.innerHTML,
+    'APPELLATE_CASE_QUERY_RESULT_PAGE',
+    callback
+  );
+  
 };
 
 // check every link in the document to see if RECAP has it
--- a/src/appellate/utils.js
+++ b/src/appellate/utils.js
@@ -65,6 +65,53 @@ let APPELLATE = {
     return /servlet\/TransportRoom/.test(url) && !/[?&]/.test(url);
   },
 
+  // Returns true if the case selection page has one row.
+  caseSelectionPageHasOneRow: () => {
+    // The Case Selection Page from Appellate shows cases that match the user's search criteria defined 
+    // on the Case Search page. This case selection can show multiple cases and has a few hidden inputs. 
+    // The csnum1 and csnum2 are two of them. These hidden input fields are populated in the following cases:
+    //
+    // - When the Case Selection Page shows one case because the user used the Case Number/Range field 
+    // to find a case by case number (csnum1 is populated, csnum2 is not populated).
+    // - When the Case Selection Page shows one or more cases because the user used the Case Number/Range 
+    // field to find cases within a range of case numbers (both inputs are populated)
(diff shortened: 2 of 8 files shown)


src/pacer.js
```diff
@@ -151,6 +151,18 @@ let PACER = {
...
+  hasTransactionReceipt(){
+    return !!$('th>font:contains("Transaction Receipt")').length;
+  },
+
+  // Returns true if the description in the transaction receipt is "Search" and
+  // the title of the page is "Select a Case"
+  isCaseQueryAdvance: function(){
+    let title = document.querySelector('#cmecfMainContent>h2')
+    return !!$('td>font:contains(Search)').length && /Select a Case/i.test(title.innerHTML);
+  },
```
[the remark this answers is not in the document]
mlissner: If you do this, you can include the comment about when people search for first/middle/last *here* instead of up above, so that whenever people use isCaseQueryAdvance, they get that explanation of why we're using a regex to look for "Select a Case".

src/content_delegate.js
```diff
@@ -361,8 +361,19 @@ ContentDelegate.prototype.handleiQuerySummaryPage = async function () {
...
     return;
   }
+  
+  if (PACER.hasTransactionReceipt() && !PACER.isCaseQueryAdvance()) {
+    // when a user tries to search a case using the first/middle/last name in the iquery form
+    // they might get redirected to a list of matching people before they reach the list of cases
+    // related to that name. Currently, we are not interested in this page where the user can
+    // choose a person.
+    //
+    // This if statement will end this function early if the user reaches this unwanted page.
```
mlissner, line 371: As explained in other comment, I suggest this goes into the isCaseQueryAdvance function.

mlissner (changes requested):
A few various comments below. 

Aside from the code, a small thing, just so you know....you say...

> will introduce
> will add
> will be added
> [...]

This is a bit unusual. Usually people say:

> introduces
> adds
> adds

In other words, usually PRs use the present tense. A small thing, but figured I better mention it.

Second:

> we created a utility function to check if the body of the page contains the transaction receipt table

It's worth mentioning a terrible thing at this point: There's a setting in PACER to turn off that table. Almost nobody has that setting on, and I think we even have a warning if you do. It's fine to assume it's there, but you reminded me that this setting existed, so I thought I'd mention it so you knew.

Anyhow, this looks generally good, ± some tweaks as below. Thank you.

src/appellate/appellate.js
```diff
@@ -35,8 +35,33 @@ AppellateDelegate.prototype.dispatchPageHandler = function () {
 
 AppellateDelegate.prototype.handleCaseSelectionPage = async function () {
   // retrieve pacer_case_id from the link related to the Case Query
-  this.pacer_case_id = APPELLATE.getCaseIdFromCaseSelection();
-  await saveCaseIdinTabStorage({ tabId: this.tabId }, this.pacer_case_id);
+  if (document.getElementsByName('csnum1')[0].value || document.getElementsByName('csnum2')[0].value){
```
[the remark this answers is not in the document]
ERosendo: I don't think Appellate PACER has a page similar to the **"Select a Person" page**. I tried to search cases using 
 the **Party/Attorney field**, the results were always shown using the **Case Selection page**.

We need this `if` statement to avoid searching the **pacer_case_id** in the **Case Selection page**( in Appellate) when there's more than one row on the page(like the last screenshot from the PR summary). We want to avoid searching the **case_id** because those pages contain results from multiple cases.

After inspecting the `body` of this page, I've found that these hidden inputs (the `csnum1` and `csnum2`) were populated in the following cases;
 
 - When the **Case Selection Page** was listing one case ( `csnum1` is populated, `csnum2` is not populated).
 - When the **Case Selection Page** was listing two cases (both hidden inputs were populated).

The lines of code inside the `if` block should be executed when **only one case** is listed on the page so we can save its pacer_case_id in the tab storage and use it in the docket reports.

Thanks to your comment, I realized that the condition of the `if` statement is wrong and should be:

``` javascript 
if (document.getElementsByName('csnum1')[0].value && !document.getElementsByName('csnum2')[0].value){

// The statements inside this block should be executed only when the page lists one case.

}
```

ERosendo:
@mlissner thanks for your comment about my PR summary. I'm going to start using the present tense from now on.

I didn't know about the setting to turn off the transaction receipt. I used the receipt because the structure of the table is consistent for Appellate and District PACER. We can use another approach to identify the **Case selection page** if the receipt table is not on the page.

We can use the following validations to identify the case selection page:

- We can check the value for `csnum1` and `csnum2` in **Appellate PACER** ( if both are `null` or `csnum2` is populated, we have more than one case listed on the page). 
- I noticed that the rows on this page in the **bankruptcy court** used two classes named `rowBackground1` and `rowBackground1` (Juriscraper uses these classes too)

I will review **District PACER** and look for an HTML element that helps us identify the page.

src/pacer.js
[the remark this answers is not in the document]
ERosendo: You're right! I will include a check for the **transaction receipt table** inside the `isCaseQueryAdvance`.

I think we can improve the `isCaseQueryAdvance` method if we include the validations I listed in my other comment so we don't have to only rely on the receipt table.

src/appellate/appellate.js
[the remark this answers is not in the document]
mlissner: Cool. In that case, I think it's worth it to wrap these checks in a function with a useful name and comment, so that it's obvious in a month what's going on here. I live in fear of node selectors that I can't explain in the future.

mlissner:
> I will review District PACER and look for an HTML element that helps us identify the page.

Well, if they have receipts turned off, we don't care too much about them — we don't support people turning off receipts, so there is no need to re-do things just for that. 

Otherwise, the rest sounds good and one other reply for you a moment ago. 

Thanks.

mlissner (commented):
Back to you!

ERosendo:
Thanks for the feedback!

These changes are already applied:

- Added a helper function in the Appellate utils file called `caseSelectionPageHasOneRow` to wrap the node selections we used in the `handleCaseSelectionPage` method of the `AppellateDelegate` class.
-  Included a check for the transaction receipt table inside the `isCaseQueryAdvance` method.
-  Moved the comment from the `ContentDelegate` class to the `isCaseQueryAdvance` method.

### Additional comments
- A new utility function to detect the **'Select a Person'** page is included. We need this helper method to avoid sending this page to the API. 
- I noticed that the extension didn't work as expected when a user reached the **Docket Report** using the links from the **Case Selection** page with multiple cases in Appellate PACER, I realized the problem was that we didn't save the **pacer_case_id** to the tab storage before the user reached the docket report ( we couldn't save it because in the previous page contained results from multiple cases). A new helper is added to fix this issue. This new utility function accesses each row of the table, retrieves the **pacer_case_id** and adds it as an URL parameter to the link that leads to the Docket report.

src/appellate/utils.js
```diff
@@ -65,6 +65,53 @@ let APPELLATE = {
...
+    //
+    // The extension is able to get the pacer_case_id and saves it to the tab storage when the Case Selection 
+    // shows only a case but this approach is not possible when multiple cases are listed so this method allows 
+    // us to support Case Selection pages with multiple cases.
+
+    document.querySelectorAll('a[href*="caseid"]').forEach((caseQueryAnchor) => {
+      let params = new URLSearchParams(caseQueryAnchor.href);
+      let caseId = params.get('caseId') || params.get('caseid')
+      // the Docket Report and the Case Query links are enclosed by the same HTML tag and the anchor for
+      // the Docket Report is the first element inside this tag so using the parentElement and the firstChild
```
mlissner, line 108: It might be helpful to show HTML in a case like this, but I get it!

src/pacer.js
```diff
@@ -151,6 +151,46 @@ let PACER = {
     return /iquery.pl/.test(url) && !/[?&]/.test(url)
   },
 
+  // Returns true if the page contains a Transaction Receipt table
+  hasTransactionReceipt: function(){
+    return !!$('th>font:contains("Transaction Receipt")').length;
+  },
```
mlissner, line 157: We do this a different way in toolbar_button.js, by checking the user's cookies. That determines something slightly different: Are receipts disabled globally? This determines if they're missing on a given page. 

Since the two things are so similar (but serve unique purposes), we should at least put them next to each other in our code base. Could you factor that one into a function called `areTransactionReceiptsDisabled`, and put it in this file?

mlissner (changes requested):
Two comments for you, then I'm pretty sure this is ready! Sorry for the delay reviewing this.

src/appellate/utils.js
ERosendo: You're right. I will add and example of the HTML structure in the comment

src/pacer.js
ERosendo: Yes! I will create this function and refactor the toolbar_button.js file.

mlissner:
Looks good. Auto-merge enabled!

mlissner (approved):

PR #296, opened 2022-12-23
feat(PDF upload): Upload pdf files from Appellate

ERosendo:
This PR adds support to upload documents from appellate PACER and inserts the banner that lets the user know a record is available in the RECAP archive. This PR resolves https://github.com/freelawproject/recap/issues/316 

- This PR updates the dipatchHandler method to include the **Single Document Page**. We use the helper method called isSingleDocumentPage that checks the name of the form available on the page.

- The Appellate class uses the same approach we use in District Court PACER to retrieve the pdf documents. In order to achieve this, this PR adds the following methods to the Appellate class:
    - **handleSingleDocumentPageView**: This method overwrites the default behavior of the submit method of the `form` element, checks the availability of the document in the **RECAP archive**, and tries to retrieve the `pacer_case_id` if its available on the page. The new submit method uses the `window.postMessage()` function instead of submitting the form.  This method also inserts a banner if the document is available in the archive.
    - **onDocumentViewSubmit**: this method receives the message from the previous one and constructs a FormData object that contains the data from the existing `form` on the page, then sends this new object using `XMLHttpRequest`.
    - **showPdfPage**: This method receives the response that the previous method created and shows the pdf document in an iframe or saves it using the `saveAs()` method.

- Creates a new file called pdf_upload.js to group utility functions related to the **Single Document page**. This utility will make the methods in the contentDelegate and appellateDelegate class shorter and more readable.

- Refactors the onDocumentViewSubmit, showPdfPage and handleSingleDocumentPageView method using the utility functions from the pdf_upload.js file.

- This PR adds a new parameter to the `httpRequest` function so we can pass the request's contentType as an argument. The current implementation of this function is not able to get the pdf files from Appellate pacer because the request is sending the default contentType (multipart/form-data) but the PACER server is expecting `application/x-www-form-urlencoded` instead.

- Creates helper functions for Appellate (`getFullPageIframe`,  `parsePdfDataFromTitle`) and District PACER( `parsePdfDataFromReceipt`, `getFullPageIframe`)

- Updates tests and the configuration file for [Karma](https://karma-runner.github.io/latest/index.html)


src/pacer.js
```diff
@@ -407,11 +407,17 @@ let PACER = {
     // this regex will match the doc_id and case_id passed as an argument in the
     // onclick event of each anchor element related to a document from Appellate Pacer.
     let doDocPost = /^return doDocPostURL\('([^']*)','([^']*)'\);/.exec(doDoc_string);
-    if (!doDocPost) {
+    let doDocPostAttachment = /^return doDocPostURL\('([^']*)'\)/.exec(doDoc_string);
```
mlissner, line 410: An example of this would be nice in the docstring, please.

src/pacer.js
```diff
@@ -443,6 +449,32 @@ let PACER = {
...
+    r.case_num = $.trim($('tr:contains(Case Number) td:nth(1)').text());
+
+    return r;
+  },
+
+  // returns HTML to create a full page iframe that loads the url passed as an argument
+  getFullPageIframe: (url) => {
+  return `<style>body { margin: 0; } iframe { border: none; }</style>
+                  <iframe src="${url}" width="100%" height="100%"></iframe>`;
+  },
```
mlissner, line 477: Weird that this is different from appellate, isn't it? Could we merge these two approaches?

mlissner: If not, we should rename this one as I said in my other comment so it's more like: `make...`

src/appellate/appellate.js
```diff
@@ -200,3 +206,179 @@ AppellateDelegate.prototype.handleAttachmentPage = async function () {
...
+
+  this.pacer_case_id = await APPELLATE.getCaseId(this.tabId, this.queryParameters, this.docId);
+  await saveCaseIdinTabStorage({ tabId: this.tabId }, this.pacer_case_id);
+
+  // When we receive the message from the above submit method, submit the form
+  // via XHR so we can get the document before the browser does.
+  window.addEventListener('message', this.onDocumentViewSubmit.bind(this), false);
+
+  let callback = $.proxy(function (api_results) {
+    console.info(`RECAP: Got results from API. Running callback on API results to ` + `insert link`);
```
mlissner, line 222: Bit funky here.

src/appellate/appellate.js
```diff
@@ -200,3 +206,179 @@ AppellateDelegate.prototype.handleAttachmentPage = async function () {
...
+    let result = api_results.results.filter((obj) => obj.pacer_doc_id == this.docId, this)[0];
+
+    if (!result) {
+      return;
+    }
+
+    let href = `https://www.courtlistener.com/${result.filepath_local}`;
```
mlissner, line 229: This is a link to a PDF, right? So:

```suggestion
    let href = `https://storage.courtlistener.com/${result.filepath_local}`;
```

src/appellate/appellate.js
```diff
@@ -200,3 +206,179 @@ AppellateDelegate.prototype.handleAttachmentPage = async function () {
...
+      console.info(`RECAP: Successfully submitted RECAP "View" button form: ${xhr.statusText}`);
+
+      const blob = new Blob([new Uint8Array(ab)], { type: type });
+      // If we got a PDF, we wrap it in a simple HTML page.  This lets us treat
+      // both cases uniformly: either way we have an HTML page with an <iframe>
+      // in it, which is handled by showPdfPage.
+      if (type === 'application/pdf') {
+        // canb and ca9 return PDFs and trigger this code path.
+        let html = APPELLATE.getFullPageIframe(URL.createObjectURL(blob));
+        this.showPdfPage(
```
mlissner, line 281: A lot of this and the stuff below is repeated, no? Why not factor it out better? Is it too different?

src/appellate/appellate.js
```diff
@@ -200,3 +206,179 @@ AppellateDelegate.prototype.handleAttachmentPage = async function () {
...
+        reader.readAsText(blob); // convert blob to HTML text
+      }
+    }.bind(this)
+  );
+};
+
+// Given the HTML for a page with an <iframe> in it, downloads the PDF
+// document in the iframe, displays it in the browser, and also
+// uploads the PDF document to RECAP.
+AppellateDelegate.prototype.showPdfPage = async function (
```
mlissner, line 319: A lot of this is repeated too, right? Could this be factored out?

src/recap.js
```diff
@@ -203,7 +203,7 @@ function Recap() {
           console.info(`RECAP: Successfully uploaded PDF: 'Success' ` +
             `with processing queue id of ${result.id}`);
           cb(result || null);
-          destroyTabStorage(cb.tab.id);
+          updateTabStorage({ [cb.tab.id]: { ['pdf_blob']: null } });
```
mlissner, line 206: I thought I'd see something explaining this, but I don't think I did. What's going on here?

mlissner (changes requested):
OK, this is going to take more than a few minutes to work on, sorry. Probably won't be worth getting it done today. 

Several comments enclosed, and a couple more oddballs below...

1. In `toolbar_button.js`, this code should be updated, right?

```
 else if (PACER.isAppellateCourt(court)) {
        // Appellate court. Show warning.
        setTitleIcon('Appellate courts are not supported', {
          '19': 'assets/images/warning-19.png',
          '38': 'assets/images/warning-38.png'
        });
```

That way we actually say it works! 

2. Really nice refactoring in content_delegate. What a relief, but it does look like there's even more of this to do (see my comments!)

Thank you. This will be an exciting one.

src/appellate/utils.js
```diff
@@ -135,6 +142,46 @@ let APPELLATE = {
...
     let [_, docId] = docString;
 
-    return docId;
+    if (docId){
+      return `${docId.slice(0, 3)}0${docId.slice(4)}`;
+    }    
+  },
+
+  // returns document data as an object
+  parsePdfDataFromTitle: (title_string) => {
```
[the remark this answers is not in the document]
ERosendo: You're right! We're not parsing pdf data. I will change the function's name.

src/appellate/utils.js
```diff
@@ -135,6 +142,46 @@ let APPELLATE = {
...
+      [, r.case_num, r.doc_num, r.att_num] = dataFromAttachment;
+    } else {
+      [, r.case_num, r.doc_num] = dataFromSingleDoc;
+      r.att_num = 0;
+    }
+    return r;
+  },
+
+  // returns HTML to create a full page iframe that loads the url passed as an argument
+  getFullPageIframe: (url) => {
```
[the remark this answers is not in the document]
ERosendo: Thanks for the suggestion. I will change the name and try to merge both helper functions.

src/pacer.js
ERosendo: They are different because I tried to use this `iframe` in Appellate, but the element didn't cover the entire viewport. 

I reviewed the issue and realized that PDF documents from District Court are wrapped in a simple `HTML page` (the content type of the response is **text/html**), but documents from Appellate are not wrapped ( the content type of the response is **application/pdf**)

The extension keeps the `HTML` elements from the response in District court and inserts the floating button that allows the user to download the file (this is the reason that we still get the whole top menu after the document is loaded), so the `iframe` inserted inherits the **CSS** styles from District court PACER.

For Appellate Court, there's no `HTML` in the response, so the extension inserts the floating button and the iframe in the root of the page, but the style in this `HTML` is not enough to fill the view area, so I created a new CSS class (`full-page-iframe`) and put the iframe inside a div element to fix the issue in Appellate Court.

I think we can use the new approach(put the iframe inside a div with the new class) for district and appellate courts, the iframe should fill the parent's view area.

mlissner: Great. I had a feeling the differences could be merged! Nice work.

src/pacer.js
ERosendo: You're right! I will add an example. Thanks.

src/appellate/appellate.js
ERosendo: You're right. Thanks!

src/appellate/appellate.js
ERosendo: The callback functions for this method are the same after we merge the two approaches to create the full-page Iframe, so we can refactor this code into a helper function to make the main method shorter and more readable.

src/appellate/appellate.js
ERosendo: You're right! I'm going to move this code into a helper function.

src/recap.js
ERosendo: The current implementation of the extension fails to identify pages if a user clicks the back button after a PDF file is uploaded from Appellate because pages like the **docket report** and the **attachment page** try to get the `pacer_case_id` from the tab storage, but the extension destroys the whole object after the upload request completed. 

I assume that the extension destroys the tab storage because it has an attribute that stores the PDF document (to share it with the background worker), but the update operation can also remove the PDF data without destroying the whole object.

ERosendo: Using the update method instead of the destroy function fixes the issue with the pages from Appellate and also removes the PDF data from the tab storage.

ERosendo:
Thanks for your comments, I've applied the suggested changes and moved some code out into a couple of helper functions to make the main methods shorter and more readable.

- The code from `toolbar_button.js` is updated and the warning about Appellate courts was removed. This fixes  https://github.com/freelawproject/recap/issues/320

- The following changes are already applied:
   
   -  A new helper method called `cleanPacerDocId` is added. This new utility function sets the fourth digit of the pacer_doc_id to 0 and has a docstring explaining the reason to do so. 
   - We merge the approaches to create the full-page iframe in a method called `makeFullPageIFrame`. The CSS class (`.appellate-full-page-iframe`) is also renamed to avoid making reference only to the appellate court.
   - The docstring in the `parseDoDocPostURL` method is updated and includes an example of the strings that are matched with the regex expressions included in this method.
   - The link for the PDFs in the RECAP archive is updated and I noticed that the code that is used to insert the banner in district courts and Appellate courts is the same, so I created a new helper method called `insertAvailableDocBanner` to avoid repeating the statements.
   - A new helper method to handle responses from the document request is added. This new utility function is called `handleDocFormResponse`. We are able to create this helper function because we merge the helper that created the full-page iframe. 
   - The code that was used in the `showPdfPage` is moved into a helper function called `showAndUploadPdf`.

### Additional comments
- I realized that the extension was not inserting a banner to create the alert in the **Case Selection Page** ( the extension inserts this banner in the **query page** in district courts). so the request to check the docket's availability is added in the `handleCaseSelectionPage `method to fix this issue.

Here's a screenshot of this page from the district and appellate court:

District Courts         |  Appellate Court
:-------------------------:|:-------------------------:
image  |  image

mlissner:
Cool! I think I could surely find more stuff to make better, but this is a huge step forward, so merging, thank you.

One comment to think about, though we don't have to fix it right away: 

> Here's a screenshot of this banner from the district and appellate court:

I appreciate how you did this the simple way by showing the banner for the appellate case, but since this page also shows the district court page, we should probably do this a bit differently. One way, for example, would be to inject [R] icons next to the two links in the results. One [R] for the appellate case, and one for the district case. That might be better, but it's certainly progress as it is!

mlissner (approved):

PR #299, opened 2022-12-29
feat(appellate): Upload case query pages to RECAP

ERosendo:
This PR adds support to upload case query pages from appellate PACER to the RECAP Archive.  This PR resolves https://github.com/freelawproject/recap/issues/317

- Adds a new upload type to the RECAP class.
- Updates the dispatch page handler in the Appellate class to include pages with the `servlet` parameter set to `'CaseQuery.jsp`'.
- Adds a new method to the Appellate class to upload the case query pages.


mlissner:
Easy enough. We'll do something with this eventually. :)

PR #301, opened 2022-12-30
fix(content.js): PACER cookie disables the extension

ERosendo:
This PR removes the checks using the PACER cookie, adds a new helper method to identify the login page, and updates the tests that use the cookie. This PR resolves https://github.com/freelawproject/recap/issues/325

I also noticed that the extension should do something when the user gets shown the `form` from the **case search page**  in the **case selection page**. The extension is returning early in cases like this but it seems like the extension is not working properly, so I think We should insert the recap email banner on this page.

Here's a gif showing how the current implementation handles the previous case:

Appellate-case-selection

Here's a gif with the changes from this PR:

Appellate-case-selection-fix


mlissner:
Seems simple enough. Fixes issues. Merging!

PR #302, opened 2022-12-30
feat(appellate): Reference district cases from Appellate

ERosendo:
this PR resolves https://github.com/freelawproject/recap/issues/21.

This PR includes the following changes:

- This PR adds two new helper methods to make **core docket numbers**. The first helper converts en & em dashes to hyphens, and it's called `normalizeDashes`, the other method makes a **core docket number** from an existing docket number, and it's called `makeDocketNumberCore`. This PR also adds tests for these methods.

- A new parameter is added to the getAvailabilityForDocket method. This new parameter is called core_docket_number and allows us to filter dockets using the `pacer_case_id `or the `docket_number_core`.

- This PR removes the `findAndStorePacerDocIds` method from the ContentDelegate class constructor. This method is asynchronous, so we should wait until it finds the doc links on the page, but we cannot use the await operator inside a constructor. This PR also updates all the invocations of the `getAvailabilityForDocket` in this class.

- Three helper functions are added to the Appellate utils.
    - `getTableWithDataFromCaseSelection`: We move some code from the `getCaseIdFromCaseSelection` method into this helper function. This function returns the table with data from the **Case Selection Page** if it's available.
    - `getDatafromDistrictLinkUrl`: this helper extracts the **court id** and the **docket number** from links to the district court available on the Case Selection Page. This utility function also computes the **core docket number**.
    - `makeRButtonForCases`: This function creates a div element that contains an anchor with the link to the district case in the RECAP archive and the RECAP icon.

- This PR moves some checks from the function that handles the response of the `getAvailabilityForDocket` method into a new helper called `handleDocketAvailabilityMessages`.

- This PR updates the `handleCaseSelectionPage` method from the `AppellateDelegate` to inject [R] icons next to the two links on the page. One [R] for the appellate case, and one for the district case ( I used the approach you suggested [here](https://github.com/freelawproject/recap-chrome/pull/296#issuecomment-1366848529)). Here's a gif with that shows the new feature:

Appellate-district-link

### Additional comments

- The response from the `getAvailabilityForDocket` might contain more than one result, in such cases the extension doesn't insert the [R] icon. Here's a gif showing this case:

Appellate-district-link-error

### DIFF
--- a/spec/RecapSpec.js
+++ b/spec/RecapSpec.js
@@ -99,8 +99,8 @@ describe('The Recap export module', function () {
       expect(jasmine.Ajax.requests.mostRecent().url).toBe(
         'https://www.courtlistener.com/api/rest/v3/dockets/' +
         '?source__in=1%2C3%2C5%2C7%2C9%2C11%2C13%2C15' +
-        '&court=canb&' +
-        'fields=absolute_url%2Cdate_modified'+
+        '&court=canb' +
+        '&fields=absolute_url%2Cdate_modified'+
         '&pacer_case_id=531316');
     });
 
--- a/spec/RecapSpec.js
+++ b/spec/RecapSpec.js
@@ -95,20 +95,20 @@ describe('The Recap export module', function () {
     it('encodes the court and caseId in the GET params', function () {
       const expectedCourt = 'canb';
       const expectedCaseId = '531316';
-      recap.getAvailabilityForDocket(expectedCourt, expectedCaseId);
+      recap.getAvailabilityForDocket(expectedCourt, expectedCaseId, null);
       expect(jasmine.Ajax.requests.mostRecent().url).toBe(
         'https://www.courtlistener.com/api/rest/v3/dockets/' +
-        '?pacer_case_id=531316' +
-        '&source__in=1%2C3%2C5%2C7%2C9%2C11%2C13%2C15' +
+        '?source__in=1%2C3%2C5%2C7%2C9%2C11%2C13%2C15' +
         '&court=canb&' +
-        'fields=absolute_url%2Cdate_modified');
+        'fields=absolute_url%2Cdate_modified'+
+        '&pacer_case_id=531316');
     });
 
     it('calls the callback with the parsed server response', function () {
       const callback = jasmine.createSpy();
       const expectedCourt = 'canb';
       const expectedCaseNum = '531316';
-      recap.getAvailabilityForDocket(expectedCourt, expectedCaseNum, callback);
+      recap.getAvailabilityForDocket(expectedCourt, expectedCaseNum, null, callback);
       jasmine.Ajax.requests.mostRecent().respondWith({
         "status": 200,
         "contentType": 'application/json',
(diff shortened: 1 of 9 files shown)


spec/RecapSpec.js
```diff
@@ -95,20 +95,20 @@ describe('The Recap export module', function () {
...
+      recap.getAvailabilityForDocket(expectedCourt, expectedCaseId, null);
       expect(jasmine.Ajax.requests.mostRecent().url).toBe(
         'https://www.courtlistener.com/api/rest/v3/dockets/' +
-        '?pacer_case_id=531316' +
-        '&source__in=1%2C3%2C5%2C7%2C9%2C11%2C13%2C15' +
+        '?source__in=1%2C3%2C5%2C7%2C9%2C11%2C13%2C15' +
         '&court=canb&' +
-        'fields=absolute_url%2Cdate_modified');
+        'fields=absolute_url%2Cdate_modified'+
+        '&pacer_case_id=531316');
```
mlissner, line 104: Nit. It's a bit sloppy how we do the ampersand here. It'd be better to always have it at the beginning:

```
'https...' +
'?xyz=foo' +
'&y=bar' +
'&z=bas';
```

src/appellate/appellate.js
```diff
@@ -72,22 +72,37 @@ AppellateDelegate.prototype.handleCaseSelectionPage = async function () {
...
           const div = document.createElement('div');
           div.classList.add('recap-banner');
           div.appendChild(recapAlertButton(this.court, this.pacer_case_id, true));
-          footer.before(recapBanner(result.results[0]));
           footer.before(div);
+
+          const rIcon = APPELLATE.makeRButtonForCases(result.results[0].absolute_url);
+          const appellateLink = anchors[0];
+          rIcon.insertAfter(appellateLink);
+        }else{
```
mlissner, line 93: I guess we should add prettier to our CI system somehow, but shouldn't this be:

```suggestion
        } else {
```

To be clear, I don't really care about this one instance, but it'd be good to be confident that our code is always clean and tidy. Some things prettier does are more important too, like adding semicolons?

src/appellate/utils.js
```diff
@@ -247,4 +262,43 @@ let APPELLATE = {
     }
     return r;
   },
+
+  // Returns an object with the court Id and docket number core extracted from a link to district court
```
mlissner, line 266: Helpful to have an example URL here.

mlissner (changes requested):
No major comments here. In general looks great and if we wait until April to merge this, it'll fix a bug that's ten years old!

Thanks.

spec/RecapSpec.js
ERosendo: You're right! 

Changed. Thanks!

src/appellate/appellate.js
ERosendo: You're right about the styling here.

src/appellate/utils.js
ERosendo: We have one! I'm going to add an example. Thanks

ERosendo:
Thanks for your comments, I've added the changes you requested.

mlissner (approved):

PR #306, opened 2023-01-06
fix(appellate): Refresh link not working in Appellate pages

ERosendo:
This PR fixes https://github.com/freelawproject/recap/issues/328

- This PR removes the mutation observer that allows us to watch for changes made to the childlist of the table element. This approach to add an `onClick` event to the **Recap Actions** button works only for district court pages. The mutation observer fails in Appellate pages because the HTML structure is different and the **RECAP actions** button is inserted in a different table.

- This PR adds an `onClick` event to the **Refresh links** button right after the extension creates it. The event handler removes the RECAP icons, shows the spinner, and invokes the `addRecapInformation` method. This new approach fixes the Refresh button for appellate pages because the extension is not relying on the mutation observer to add the onClick event.

- We add checks in the appellate class to avoid duplicating the RECAP actions button.

Here are some gifs of the refresh link button working in **District** and **Appellate** courts:

**Appellate PACER**

refresh-links-appellate

**District PACER**

refresh-links-district

### DIFF
--- a/src/action_button.js
+++ b/src/action_button.js
@@ -72,6 +72,20 @@ const recapDropdownMenu = (court, pacerCaseId) => {
   checkDoc.setAttribute('id', 'refresh-recap-links');
   checkDoc.setAttribute('role', 'button');
 
+  checkDoc.addEventListener('click', () => {
+    let links = document.querySelectorAll('.recap-inline, .recap-inline-appellate');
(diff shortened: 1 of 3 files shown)


mlissner:
As the zen of Python says, simple is better than complex! Nice.

PR #308, opened 2023-01-07
fix(attachment page): Add missing attachment number in Appellate

ERosendo:
This PR resolves https://github.com/freelawproject/recap/issues/327.

This PR includes the following changes:

- This PR will update the if statement in the `getAttachmentNumberFromAnchor` method to avoid looking for an attachment number if the row has 3 or fewer children nodes. The current implementation uses the `<` which fails if the row that contains the doc links has exactly 3 children nodes ( The table from the docket report has exactly 3 data cell elements but we don't want to search the attachment number in that table).

- This PR adds the `attachment number` to the doc links URL if it's found on the page and also makes sure that the `caseId `is always added to the href attribute.

- We add a new if statement to check the queryparameters

### DIFF
--- a/src/appellate/appellate.js
+++ b/src/appellate/appellate.js
@@ -370,8 +370,8 @@ AppellateDelegate.prototype.onDocumentViewSubmit = function (event) {
   let title = document.querySelectorAll('strong')[1].innerHTML;
   let dataFromTitle = APPELLATE.parseReceiptPageTitle(title);
 
-  if (dataFromTitle.att_number == 0 && this.queryParameters.get('attNum')){
-    dataFromTitle.att_number = this.queryParameters.get('attNum')
+  if (dataFromTitle.att_number == 0 && this.queryParameters.get('recapAttNum')){
+    dataFromTitle.att_number = this.queryParameters.get('recapAttNum')
   }
 
   if (!dataFromTitle) {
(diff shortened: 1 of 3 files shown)


src/appellate/appellate.js
```diff
@@ -369,6 +369,11 @@ AppellateDelegate.prototype.onDocumentViewSubmit = function (event) {
 
   let title = document.querySelectorAll('strong')[1].innerHTML;
   let dataFromTitle = APPELLATE.parseReceiptPageTitle(title);
+
+  if (dataFromTitle.att_number == 0 && this.queryParameters.get('attNum')){
```
mlissner, line 373: Hm, a thought. If we're adding query parameters just for our own sake, should we namespace them? In other words, could we make any parameter we use start with recap, say? `recapAttNum`?

src/pacer.js
```diff
@@ -644,7 +644,7 @@ let PACER = {
     //  attachment number of the document.
 
     let row = anchor.parentNode.parentNode;
-    if (row.childElementCount < 3) {
+    if (row.childElementCount <= 3) {
```
mlissner, line 647: Should we add a comment here to explain the different number of elements that are possible?

mlissner (commented):
Couple quick suggestions for you, Eduardo!

src/appellate/appellate.js
ERosendo: We can namespace the parameters. That's a great idea, I will change The set() method of the [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) interface.

src/pacer.js
ERosendo: Sure.

PR #311, opened 2023-01-10
feat(district): Add button to autofill docket query form

ERosendo:
This PR resolves https://github.com/freelawproject/recap/issues/190.

- A new field is added to the `getAvailabilityForDocket` request. This PR adds the `date_last_filing` field to the list of dynamically selected fields.

- This PR adds two helper functions in the utils.js file. 

    - `pacerDateFormat`: takes a date and returns it in 'MM/DD/YYYY' format
    - `recapAddLatestFilingButton`: creates the HTML button that is inserted on the page and also creates an onClick event listener for this object. This utility function uses the `date_last_filing` from the response of `getAvailabilityForDocket` request to create a data attribute in the button.

- This PR adds CSS classes to change the styling of the new button.

- The helper method called `removeBanners` is updated to also remove the new button. 

- This PR adds the helper method `recapAddLatestFilingButton` to the Docket Query handler to create the filing button next to the date input field of the form.

Here's a gif showing the feature:

refresh-button-district

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -215,6 +215,11 @@ const removeFilingState = () => {
   document.cookie = "isFilingAccount=false;path=/;domain=.uscourts.gov";
 }
 
+//takes a date in 'YYYY-MM-DD' format returns it in 'MM/DD/YYYY' format
+function pacerDateFormat(date) {
+  return date.replace(/(\d+)-(\d+)-(\d+)/, "$2/$3/$1");
+}
+
 // Default settings for any jquery $.ajax call.
 $.ajaxSetup({
   // The dataType parameter is a security measure requested by Opera code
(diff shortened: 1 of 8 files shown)


src/utils.js
```diff
@@ -215,6 +215,11 @@ const removeFilingState = () => {
   document.cookie = "isFilingAccount=false;path=/;domain=.uscourts.gov";
 }
 
+//takes a date in 'YYYY-MM-DD' format returns it in 'MM/DD/YYYY' format
```
mlissner, line 218:
```suggestion
// converts an ISO-8601 date str to 'MM/DD/YYYY' format
```

src/utils.js
```diff
@@ -283,7 +288,39 @@ const recapAlertButton = (court, pacerCaseId, isActive) => {
...
 
+// Creates an anchor element to autofill the Docket Query form
+const recapAddLatestFilingButton = (result) =>{
+  
+  let date = result.date_last_filing
+  let formatted_date = pacerDateFormat(date)
+
+  const anchor = document.createElement('a');
+  anchor.classList.add("recap-filing-button");
+  anchor.title = 'This will purchase filings since the latest we have on RECAP, omitting parties and member cases.'
```
mlissner, line 299:
```suggestion
  anchor.title = 'Autofill the form to get the latest content not yet in RECAP, omitting parties and member cases.'
```

src/utils.js
```diff
@@ -283,7 +288,39 @@ const recapAlertButton = (court, pacerCaseId, isActive) => {
   return anchor;
 };
 
+// Creates an anchor element to autofill the Docket Query form
+const recapAddLatestFilingButton = (result) =>{
+  
+  let date = result.date_last_filing
```
mlissner, line 294: When we don't have a docket or this field is null, does everything work properly?

mlissner (changes requested):
A few q's/comments, but approach looks sound, thank you!

src/utils.js
ERosendo: If We don't have a docket, the extension doesn't insert the button. but It is not working properly if this field is null. 

What should We do in such a case? avoid inserting the button or insert a button that only unchecks "Parties and counsel"?

let me know what you think.

mlissner: I hadn't mentioned it, but the ideal version of this button checks if we have party/counsel info and only unchecks if we do. That's hard, but at least in this case if we don't have the docket, we can just not show the button, and that's pretty close.

ERosendo: Thanks for the explanation. I will make sure We don't show the button if we find one of these cases and I'll include tests to avoid regressions.

mlissner (approved):

mlissner:
Awesome. If this passes your QA, let's deploy today!

PR #315, opened 2023-01-26
fix(appellate): Avoid uploading case summary with the wrong case_id

ERosendo:
This PR fixes https://github.com/freelawproject/recap/issues/330.

-  This commit adds a new helper function called `getDocketNumber`. This new helper uses different approaches to get the docket number from the appellate pages.

- A new URL query parameter is added to the anchor's href attribute. This new query parameter is called `recapCaseNum`. This new parameter will be used as the last resource to get the docket number from an appellate page.

- This PR updates the methods `handleCaseSelectionPage`, `handleCaseQueryPage`, `attachRecapLinksToEligibleDocs`, `handleDocketDisplayPage` and `handleSingleDocumentPageView` from the AppellateDelegate class to save the docket number to the tab storage.

- Update the helper method called `getCaseId` to check the docket number of the appellate page before using the `pacer_case_id` saved in the local storage.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,8 +11,8 @@ Changes:
  - None yet
 
 Fixes:
- - None yet
- - 
+ - More robust uploading of case summary pages from appellate ([#330](https://github.com/freelawproject/recap/issues/330))
+  
 For developers:
  - Nothing yet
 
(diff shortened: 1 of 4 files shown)


mlissner:
Looks good, I think. We'll see if it causes any trouble!

PR #322, opened 2023-02-07
Fix broken icon in iOS

ERosendo:
This PR fixes issues that We found in Safari and resolves https://github.com/freelawproject/recap/issues/332. 

- We use a new approach to create the content of the banners. The current implementation of the extension creates a banner with a broken icon in safari so in order to fix that We're adding a helper function called `makeMessageForBanners`, this utility method creates the content of the banner and returns a `div` element.

- This PR updates the file called `style-ios.css`. This file has all the CSS classes that are used to build the extension that is going to be installed on iPhones.

- We're now injecting the content scripts using the background.js file to avoid **duplication**. This issue seems to be pretty common because Safari injects scripts into the **top-level page** and any **subpages** with HTML sources. The [docs ](https://developer.apple.com/documentation/safariservices/safari_app_extensions/using_injected_style_sheets_and_scripts?language=objc) for Safari extension developers say that we should not assume that there’s only one instance of your script per browser tab. We're making the following changes to accomplish the [programmatically injection](https://developer.chrome.com/docs/extensions/mv2/content_scripts/#programmatic
) of the scripts:

    - We add the [activeTab](https://developer.chrome.com/docs/extensions/mv2/manifest/activeTab/) permission in the manifest.json file
    
    - We're removing all the js files that matches **URLs** containing the string **uscourts.gov** from the `manifest.json` and creating a new list that contains all the files listed in the manifest.json in the background page. This new list is called `CONTENT_SCRIPT_FILES` .
    
    - We're using the **onUpdated** event of the chrome.tabs API to start injecting the js files to the page.
   
    - The background script sends a message to the tab checking if the content scripts are already inserted, if the tab doesn't send a response to the background, It will start injecting the content scripts. We need this extra step before injecting the js files because the onUpdated event is fired several times when a user navigates to a new URL in a Tab and We just want to use the event the first time that's fired.

### DIFF
--- a/src/assets/css/style.css
+++ b/src/assets/css/style.css
@@ -170,6 +170,10 @@ footer #version {
   color: #aaa;
 }
 
+.d-inline-flex{
+  display: inline-flex !important;
+}
+
 .recap-email-banner-full{
   font-family: helvetica, arial, serif;
   font-size: 13px;
@@ -508,15 +512,11 @@ footer #version {
   padding-top: 2px;
 }
 
-.recap-banner-inner-div{
-  display: inline-flex;
-}
-
-.recap-banner-inner-div > p{
+.banner-message > p {
   margin: 0px;
   padding: 0px 2px;
 }
 
-.recap-banner-inner-div > img{
+.banner-message > img{
   margin-right: 0.25rem;
 }
--- a/src/assets/css/style.css
+++ b/src/assets/css/style.css
@@ -507,3 +507,16 @@ footer #version {
 .recap-filing-button > img{
   padding-top: 2px;
 }
+
+.recap-banner-inner-div{
+  display: inline-flex;
+}
+
+.recap-banner-inner-div > p{
+  margin: 0px;
+  padding: 0px 2px;
+}
+
+.recap-banner-inner-div > img{
+  margin-right: 0.25rem;
+}
--- a/src/utils.js
+++ b/src/utils.js
@@ -364,6 +364,9 @@ const recapBanner = (result) => {
   let message = `View and Search this docket as of ${time.outerHTML} for free from RECAP`;
   let innerDiv = makeMessageForBanners(message)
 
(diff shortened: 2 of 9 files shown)


src/utils.js
```diff
@@ -282,9 +282,21 @@ const recapAlertButton = (court, pacerCaseId, isActive) => {
...
+  p.innerHTML = text
+  
+  const innerDiv = document.createElement('div')
+  innerDiv.classList.add("recap-banner-inner-div");
+  
+  innerDiv.appendChild(img)
+  innerDiv.appendChild(p)
+
+  anchor.append(innerDiv)
+
```
mlissner, line 299: Feels like this (and possibly more?) could be in a little utility function.

src/assets/css/style.css
```diff
@@ -506,4 +506,17 @@ footer #version {
 
 .recap-filing-button > img{
   padding-top: 2px;
+}
+
+.recap-banner-inner-div{
```
mlissner, line 511: Classes shouldn't normally be named after elements. Maybe use a better rule to get these to apply like you want?

mlissner (changes requested):
Couple little things, but if you say this works, great.

src/utils.js
ERosendo: You're right! I've moved this code to an utility function.

src/assets/css/style.css
ERosendo: got it! I've change the name of this class.

mlissner:
I'm going to duck out of this one. @albertisfu could you please review and I'll trust whatever you guys came up with.

albertisfu (approved):
Thanks @ERosendo this looks good to me. I tested all the functions in safari for desktop and safari for iOS and it's working very well, so this can be merged and I'll continue with the safari release.

mlissner (approved):

PR #324, opened 2023-04-20
appellate/utils.js: Use URLSearchParams() properly

johnhawkinson:
CA9 and CA2 both use ACMS for some dockets (**Edit: not just** immigration), which are hosted on an external website with a different URL form. So we have this sort of thing:

Screen Shot 2023-04-19 at 22 42 47

with:

```html
<td>
  <a href="https://ca2-showdoc.azurewebsites.us/23-6359" target="_blank">
    23-6359
  </a>
  <br>
  <a href="TransportRoom?servlet=CaseQuery.jsp&amp;cnthd=2026821867&amp;caseid=1002765&amp;csnum1=23-6359&amp;shorttitle=United+States+of+America+v.+Conde+%28Kone%29">
    United States of America v. Conde (Kone)
  </a>
  &nbsp;
</td>
```

And the current code breaks that first `<a>` link by appending a bare `&caseId` to it, which is improper since there is no preexisting `?`, and results in a URL that doesn't work and breaks appellate PACER. Doh!

Fix this and a few other instances of URL handling in `utils.js`, thusly. This code is untested:

<hr>

Don't feed the URLSearchParams() constructor an entire URL, it takes only a query string. Feeding it a full URL causes it to misparse the first query string, if any, merging it with the URL. In practice this was not an actual problem, but could have been.

Similarly, don't set a URL search parameter by concatenating "&caseid=xyz"; this fails when there is no preexisting search string:(i.e. no ?foobar), and causes the RECAP extension to break URLs. Instead, use the URL+URLSearchParamers abstractions.

This has the side effect of converting relative URLs to absolute.  I don't think this is an actual problem?

Lastly, don't parse a query string by hand with split('?') -- use the URL constructor to do it.

### DIFF
--- a/src/appellate/utils.js
+++ b/src/appellate/utils.js
@@ -159,9 +159,9 @@ let APPELLATE = {
       // attribute allow us to get the desired HTML element.
       let caseSummaryAnchor = caseQueryAnchor.parentElement.firstChild;
       // This has the side effect of making this URL absolute, when it may have started out relative.
-      summaryUrl = new URL(window.location, caseSummaryAnchor.href);
+      let summaryUrl = new URL(window.location, caseSummaryAnchor.href);
       summaryUrl.searchParams.set('caseId', caseId);
-      caseSummaryAnchor.setAttribute('href', summaryURL);
+      caseSummaryAnchor.setAttribute('href', summaryUrl);
     });
   },
 
--- a/src/appellate/utils.js
+++ b/src/appellate/utils.js
@@ -151,13 +151,17 @@ let APPELLATE = {
     // us to support Case Selection pages with multiple cases.
 
     document.querySelectorAll('a[href*="caseid"]').forEach((caseQueryAnchor) => {
-      let params = new URLSearchParams(caseQueryAnchor.href);
-      let caseId = params.get('caseId') || params.get('caseid');
+      let queryUrl = new URL(window.location, caseQueryAnchor.href);
+      let queryParams = queryUrl.searchParams;
+      let caseId = queryParams.get('caseId') || queryParams.get('caseid');
       // the Docket Report and the Case Query links are enclosed by the same HTML tag and the anchor for
       // the Docket Report is the first element inside this tag so using the parentElement and the firstChild
       // attribute allow us to get the desired HTML element.
       let caseSummaryAnchor = caseQueryAnchor.parentElement.firstChild;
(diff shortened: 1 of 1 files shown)


ERosendo:
@johnhawkinson Thanks for your comments. This code looks good. I tested your changes and it seems to work well with absolute and relative URLs. 

On the other hand, I realized the current implementation of the extension won't upload data from dockets like these because they're hosted on another website with an URL that doesn't match the pattern We use in the `manifest.json` file. We need to add a pattern that matches `https://ca2-showdoc.azurewebsites.us/` to the `content_scripts` array in the `manifest.json` file.

@johnhawkinson do you know if CA9 uses the same **root** domain to host the dockets?

johnhawkinson:
whoops sorry about the constructor argument ordering.

My focus was on making sure RECAP didn't break PACER rather than making
sure RECAP actually worked. It would certainly not surprise me if there are
other reasons ACMS does not work right.

I haven't encountered any ca9 ACMS cases recently but i'm sure you could
find some easily enough, although the query does cost 10 cents. The ca2
query was just for cases files in the past 2 days.

That said, https://ca9-showdoc.azurewebsites.us/ does exist and I can't
imagine ca9 is is using the ca2 website, esp. since ca9 deployed ACMS
first. Wildcarding all of azurewebsites.us seems clearly wrong and even
wildcarding ca*-showdoc.azurewebsites.us feels a little risky but is
probably the right thing for now. Or just listing all 13 by hand.

Something should probably be set up to monitor the URLs since they seem
sure to change in the future, e.g. to uscourts.gov.

mlissner:
I'm not thrilled about tweaking the manifest because whenever we do a bunch of people uninstall, but I guess they're not important users anyway, so @ERosendo, will you please tweak the manifest to add the needed domains? I think just explicitly spelling them out is best, rather than using a wildcard.

Beyond that, can you please check if things work properly on that website? I suspect they won't for some reason. If they don't, we can start by un-breaking RECAP in this PR, and then do another one later to add support for these websites.

mlissner:
> I don't know what the risk is of a malicious person coming around and creating an unauthorized Fifth Circuit ACMS website?

I mean, most extensions just have a big fat wildcard in their manifest, so I'm not *super* worried. I guess somebody could:

1. Make a fake ca5 ACMS
2. Somehow trick people into going to it
3. Get us to upload bad info (zip bomb? JS in the HTML?) to the archive

But that seems like a lot of work for not much benefit. We already sanitize everything we get.

PR #325, opened 2023-04-21
Be clear when we modify links in a webpage, by adding two markers.

johnhawkinson:
Be clear when we modify links in a webpage, by adding two markers.

Specifically, change two attributes:
* Add a `data-recap` attribute explaining that it was modified, and
* Append `recap_modified` to the CSS class list.

Resulted from a Slack discussion this morning in light of #324 where it was not obvious upon inspection that the RECAP extension was messing with the page and breaking the URL, and it would have aided debugging to have done so.

I don't quite understand the test infrastructure and where to add a test to check this stuff -- it seems like there's zero test coverage for `addCaseIdToDocketSummaryLink()` and I don't know quite where to begin, so no tests.

As a prereq, consistently use the `.dataset` API rather than rolling our own.

### DIFF
--- a/src/appellate/utils.js
+++ b/src/appellate/utils.js
@@ -272,14 +272,14 @@ let APPELLATE = {
       // store extra information on anchors to use it while handling the onClick listener
       let docId = PACER.getDocumentIdFromUrl(clonedNode.href);
 
-      clonedNode.setAttribute('data-pacer_doc_id', docId);
+      clonedNode.dataset.pacerDocId = docId;
       if (doDoc && doDoc.doc_id) {
-        clonedNode.setAttribute('data-pacer_dls_id', doDoc.doc_id);
+        clonedNode.dataset.pacerDlsId = doDoc.doc_id;
       }
-      clonedNode.setAttribute('data-pacer_case_id', pacerCaseId);
-      clonedNode.setAttribute('data-pacer_tab_id', tabId);
-      clonedNode.setAttribute('data-document_number', docNum ? docNum : docId);
-      clonedNode.setAttribute('data-attachment_number', attNumber);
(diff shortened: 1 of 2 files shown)


src/appellate/utils.js
```diff
@@ -272,14 +274,14 @@ let APPELLATE = {
...
+        clonedNode.dataset.pacerDlsId = doDoc.doc_id;
       }
-      clonedNode.setAttribute('data-pacer_case_id', pacerCaseId);
-      clonedNode.setAttribute('data-pacer_tab_id', tabId);
-      clonedNode.setAttribute('data-document_number', docNum ? docNum : docId);
-      clonedNode.setAttribute('data-attachment_number', attNumber);
+      clonedNode.dataset.pacerCaseId = pacerCaseId;
+      clonedNode.dataset.pacerTabId = tabId;
+      clonedNode.dataset.documentNumber = docNum ? docNum : docId;
+      clonedNode.dataset.attachmentNumber = attNumber;
```
mlissner, line 284: Is the changing of snake_case to camelCase OK here? Looks scary, but I imagine you wouldn't do this unless it was equivalent?

johnhawkinson: Yeah, it's how the dataset API is defined, cf. https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset#name_conversion
I guess it's accomplished with a snazzy DOMStringMap thing. I don't disagree that it is potentially confusing, yet it still seemed better to use the API for writing this data rather than hand-rolling it. Shrug.

mlissner: I get it. I guess HTML is better with dashes and JS is better with camels. Great. I'll merge.

mlissner:
And thank you!

PR #326, opened 2023-04-29
Add support for audio file PDFs and alternate transcript receipt format

ttys0dev:
Fixes [#244](https://github.com/freelawproject/recap/issues/244).

### DIFF
--- a/spec/PacerSpec.js
+++ b/spec/PacerSpec.js
@@ -4,6 +4,7 @@ describe('The PACER module', function () {
   const noTrailingSlashUrl = 'https://ecf.canb.uscourts.gov';
   const docketReportURL = 'https://ecf.canb.uscourts.gov/cgi-bin/DktRpt.pl';
   const singleDocUrl = 'https://ecf.canb.uscourts.gov/doc1/034031424909';
+  const singleDocAudioUrl = 'https://ecf.canb.audio.uscourts.gov/doc1/034031424909';
   const docketQueryUrl = 'https://ecf.canb.uscourts.gov/cgi-bin/' + 'HistDocQry.pl?531316';
   const docketQueryUrlFromAppellate =
     'https://ecf.canb.uscourts.gov/cgi-bin/' + 'DktRpt.pl?caseNumber=2:16-cv-01129-RFB-DJA';
@@ -36,6 +37,10 @@ describe('The PACER module', function () {
 
     it('matches a valid single doc URL', function () {
       expect(PACER.getCourtFromUrl(singleDocUrl)).toBe('canb');
+    });
+
+    it('matches a valid single doc audio URL', function () {
+      expect(PACER.getCourtFromUrl(singleDocAudioUrl)).toBe('canb');
     });
 
     it('ignores patent nonsense', function () {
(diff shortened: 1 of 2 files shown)


spec/PacerSpec.js
```diff
@@ -4,6 +4,7 @@ describe('The PACER module', function () {
   const noTrailingSlashUrl = 'https://ecf.canb.uscourts.gov';
   const docketReportURL = 'https://ecf.canb.uscourts.gov/cgi-bin/DktRpt.pl';
   const singleDocUrl = 'https://ecf.canb.uscourts.gov/doc1/034031424909';
+  const singleDocAudioUrl = 'https://ecf.canb.audio.uscourts.gov/doc1/034031424909';
```
ttys0dev, line 7: Some audio file `View Document` forms(such as those in the `paed` court) contain this url format, the domain doesn't actually resolve or get used directly it would appear, however we need to support parsing these for the extension to work properly.

src/pacer.js
```diff
@@ -564,9 +566,17 @@ let PACER = {
     //  - doc_number
     //  - att_number
     
+    let audio_string = $('td:contains(AUDIO)').text();
     let image_string = $('td:contains(Image)').text();
+    let transcript_string = $('td:contains(TRANSCRIPT:)').text();
```
ttys0dev, line 571: Some transcript files(such as those in `laeb`) use a `TRANSCRIPT:` prefix instead of a `Image` prefix.

mlissner:
Wow, busy weekend. It looks like this one is ready too?

ttys0dev:
> It looks like this one is ready too?

Yep.

mlissner:
Great. I'm putting it on the top of @ERosendo's backlog to review, but he's on a high priority project right now so he might be a moment until he can come up for air. But we do try to prioritize PRs, so hopefully he can switch to it once his current thing is In Review.

ERosendo:
Sure!

src/pacer.js
```diff
@@ -586,9 +589,17 @@ let PACER = {
     //  - doc_number
     //  - att_number
     
+    let audio_string = $('td:contains(AUDIO)').text();
```
ERosendo, line 592: We should refactor this method. I think we can use short-circuit evaluation to avoid repeating the regex search. 

We can do something like this:

```javascript
let audio_string = $('td:contains(AUDIO)').text();
let image_string = $('td:contains(Image)').text();
let transcript_string = $('td:contains(TRANSCRIPT)').text();

let receipt_description = image_string || audio_string || transcript_string;
let regex = /(\d+)-(\d+)/;
let matches = regex.exec(receipt_description);
if (!matches) {
  return null;
}
```

ERosendo (changes requested):
This looks good. I tested your code by uploading a few audio files from [4:17-cr-03038](https://www.courtlistener.com/docket/6357941/united-states-v-parsons/) and It worked great. 

Just one comment.

Thanks!

src/pacer.js
ttys0dev: applied as suggested

ERosendo: thanks!

ERosendo (approved):

PR #327, opened 2023-06-07
feat(pacer): Filter out show_doc links

ERosendo:
This commit fixes cases iii and iv from https://github.com/freelawproject/recap/issues/341.

- Adds a new utility function that only matches `doc1` URLs. This new helper function allows us to filter out all the `show_doc.pl` links on the reports before the extension tries to extract the `pacer_doc_id`.

- Add tests for the method called` isDoc1Url`.

- Updates the docstring for the `isDocumentUrl` method.


mlissner (approved):
LGTM, thank you!

PR #329, opened 2023-06-08
feat(content.js): Tweak logic to get the pacer_case_id from input

ERosendo:
This commit tweaks the logic to check the value of the hidden input with the id 'all_case_ids' before updating the `pacer_case_id`.

We checked this input in https://github.com/freelawproject/recap-chrome/pull/273 and found that it was empty when there wasn't a cookie for the case number ( here's a comment about it: https://github.com/freelawproject/recap-chrome/pull/273#pullrequestreview-1184750272) but I noticed that the input could also use `value='0'`, so the extension updates the `pacer_case_id`, tries to lookup a case using the API and it also creates a banner in the **Docket Query Page** linked to this [case](https://www.courtlistener.com/docket/66992224/united-states-v-sandoval/) but the user hasn't search any case.

Here's a screenshot of this bug:

image

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,8 @@ Changes:
 Fixes:
  - Support for Safari on macOS and iPhone([#332](https://github.com/freelawproject/recap/issues/332)).
  - Better document link filtering on district and appellate court reports([#341](https://github.com/freelawproject/recap/issues/341)).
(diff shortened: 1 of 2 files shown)


mlissner (approved):
Thanks!

PR #332, opened 2023-07-01
feat(appellate): Add logic to get doc_id from the dls_id parameter

ERosendo:
This PR fixes https://github.com/freelawproject/recap/issues/340


mlissner:
Simple enough. Includes change notes and tests. No complaints here!

PR #333, opened 2023-07-02
feat(appellate): Add banners to the report filter page

ERosendo:
This PR fixes https://github.com/freelawproject/recap/issues/335.

Here's a screenshot of the banners on the filter page:

image


mlissner:
Almost too easy, but we do need release notes. Thank you!

PR #335, opened 2023-07-10
Improve Document Selection Menu page matching

ttys0dev:
The existing match conditions for attachment pages doesn't seem to be reliable(ie for example on `cacd`), lets just add a match condition for the "Document Selection Menu" text at the top as well which seems to fix these issues.

### DIFF
--- a/src/pacer.js
+++ b/src/pacer.js
@@ -317,9 +317,11 @@ let PACER = {
     let mainContent = document.getElementById("cmecfMainContent");
     // End this function early if we're on a management PACER page
     if (!mainContent){ return false }
-    let bottomNote = mainContent.lastChild.textContent.includes('view each document individually')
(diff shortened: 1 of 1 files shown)


ERosendo:
We could try, but it's the JS code from PACER that's creating those errors, not the extension. 

The extension works properly using this approach to identify attachment pages with this issue. It attaches the Recap icon to documents that we have in the RECAP archive and it can upload new ones. I uploaded attachment 3 of [entry #1 ](https://ecf.cacd.uscourts.gov/doc1/031040373760?caseid=890735)from [2:23-cv-05413](https://www.courtlistener.com/docket/67564549/smith-v-american-airlines-inc/).

mlissner:
Got it, thanks. Merging, but we need release notes too. :)

PR #336, opened 2023-07-11
fix(appellate): Normalize document number and dls_ids

ERosendo:
This PR adds logic to normalize the document number when the appellate court uses the pacer_doc_id instead of the regular docket entry numbering.

### DIFF
--- a/spec/AppellateSpec.js
+++ b/spec/AppellateSpec.js
@@ -13,8 +13,8 @@ describe('The Appellate module', function () {
       onClick: "return doDocPostURL('009031927529','290338');",
     },
     {
-      href: 'https://ecf.ca9.uscourts.gov/docs1/009031956734',
-      onClick: "return doDocPostURL('009031956734','290338');",
+      href: 'https://ecf.ca9.uscourts.gov/docs1/009131956734',
+      onClick: "return doDocPostURL('009131956734','290338');",
     },
   ];
   const searchParamsWithCaseId = new URLSearchParams('servlet=DocketReportFilter.jsp&caseId=318547');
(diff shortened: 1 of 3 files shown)


mlissner:
Looks good, but again, no release notes. :)

mlissner:
Does this fix https://github.com/freelawproject/courtlistener/issues/2877, or is more needed?

ERosendo:
This PR won't fix https://github.com/freelawproject/courtlistener/issues/2877.  

While we were reviewing the issue we noticed the main problem was that we were updating the `document number` of the `RECAPDocument` records with something that looks like a non-normalized doc_id, but this document number is interesting because it doesn't have leading zeros, so I checked the code that handles the extraction of this number from PACER and noticed we don't remove the zeros then I checked the request that uploads the PDF to the RECAP archive and realized the API was the one removing the zeros. Here's an example 

- Request 
```
court: ca8
pacer_case_id: 105396
pacer_doc id: 00804758071
document number: 00814758071
attachment number: 2 
filepath_local: (binary)
upload type: 3 
debug: false
```

- Response 
```
attachment number: 2 
court: "ca8" 
date created: "2023-07-10718:42:11.438895-07:00" 
date _modified: "2023-07-10718:42:11.438909-07:00"
debug: false docket: null 
docket_entry: null 
document _number: 814758071
error message:
id: 9130623 
pacer case id: "105396" 
pacer_doc_id: "00804758071" 
recap document: null 
status: 1
upload type: 3
```

We found the problem with the API is that the `ProcessingQueue` model uses a `BigIntegerField` to store the `document number`( [here's](https://github.com/freelawproject/courtlistener/blob/21dfc5a1e244533b5ff1bc0aecdc489d3841c1bb/cl/recap/models.py#L127-L131) the code) and that's why it's removing the zeros. so even if the extension is sending normalized document numbers like `"00804758071"`, the `ProcessingQueue` record will remove the leading zeros and store something like `804758071` and for that reason, Alberto suggested we could check if the PDF upload belongs to an attachment page and it's using the `pacer_doc_id` as the `document_number`, so we can avoid updating the `document number` we already have.

PR #338, opened 2023-07-13
fix(appellate): Remove the logic to normalize the pacer_dls_id

ERosendo:
While going through the QA steps, I realized the extension failed to upload free documents. After debugging the code, I noticed we're using the pacer_dls_id attribute in the [onClickEventHandlerForDocLinks](https://github.com/freelawproject/recap-chrome/blob/main/src/appellate/utils.js#L210) to create the request that allows us to check whether the doc link returns an HTML page (the view document page) or a pdf file (when it's free). sending the normalized version of the dls_id always returns an HTML page even when the doc is an opinion (which is a free doc).

This PR removes the logic to normalize the `dls_id.`


mlissner:
Seems OK, but tests are failing. :)

ERosendo:
@mlissner it is ready now. The new action to check the release notes is working properly :).

PR #344, opened 2023-09-27
README: Clarify eslint is aspirational

johnhawkinson:
It's been FOUR YEARS since cb38fb52 and the tree is not eslint compliant. Don't gaslight developers into thinking they're doing it wrong when the reality is that style and linting claim is solely aspirational.

I do not appreciate wasting my time trying to figure out what is wrong with my development environment because of stuff like this:

```
jhawk@lrr recap-chrome % eslint src| tail -4

✖ 16159 problems (16159 errors, 0 warnings)
  14851 errors and 0 warnings potentially fixable with the `--fix` option.

```


mlissner:
Setting automerge. Sorry for the confusion and trouble and thanks for fixing it for future developers.

PR #345, opened 2023-09-27
manifest.json: More COA ACMS websites

johnhawkinson:
ACMS docket report webpages return Javascript that sets the `_models` attribute of the `window` global object to structued data, which include a reference to API urls, e.g.:

  >> window._model.AppSettings.ApiUrl
  "https://ca9-showdocservices.azurewebsites.us/api"

Consequently, include the ca*-showdocservices websites in our permission list, again out of an abundance of caution.


mlissner:
Thanks. We're doing the right thing here by being explicit, but I worry that a really long list of domains is scarier to many users than a wildcard. Would it work to do something like `ca*.azurewebsites.us`? I dunno. It might be less scary to people.

johnhawkinson:
In fact it would **not work**.
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Match_patterns#host

<hr>
<div class="section-content"><p>The <em>host</em> component may take one of three forms:</p>
| Form | Matches
| -- | --
| * | Any host.
| *. followed by part of the hostname. | The given host and any of its subdomains.
| A complete hostname, without wildcards. | Only the given host.
</figure>
<p><em>host</em> must not include a port number.</p>
<p><em>host</em> is optional only if the <em>scheme</em> is "file".</p>
<b><p>Note that the wildcard may only appear at the start.</p></b></div>
<hr>

<b>emphasis</b> mine

mlissner:
Well that settles that. Thanks.

PR #346, opened 2023-09-27
feat(toolbar): Show warning sign in ACMS websites

ERosendo:
This PR adds a helper method to identify ACMS URLs and logic to show a warning sign in the toolbar icon when users are on those new websites.

Here's a screenshot of the toolbar icon:

image

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,7 +5,7 @@
 The following changes are not yet released, but are code complete:
 
 Features:
- - None yet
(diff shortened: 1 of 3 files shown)


mlissner:
Thanks!

daniel-walker:
Are they're plans for support ACMS ?

mlissner:
Yeah, we have some code for it, but it's not over the finish line yet. Hopefully soon!

daniel-walker:
I'm finding this warning on an increasing number of cases in the 9th circuit court of appeals. It appears to be most new cases, and court listener doesn't cache anything in those cases.

mlissner:
Yes, all new CA9 cases are in ACMS.

PR #348, opened 2023-11-15
fix(content_delegate): Fix zip upload in district courts

ERosendo:
This PR fixes https://github.com/freelawproject/recap/issues/352 by tweaking the `extractUrl` method to retrieve the URL from the `script` tag. During the debugging process, I noticed the absence of a message indicating that the extension was uploading the zip file. To address this, I added a loading message on the page to inform users that the extension is working on the upload task.

Here's a GIF showing the new message:

upload zip 2

### Additional comments

I couldn't reproduce the issue on Windows, it seems specific to macOS.

### DIFF
--- a/src/content_delegate.js
+++ b/src/content_delegate.js
@@ -595,9 +595,7 @@ ContentDelegate.prototype.onDownloadAllSubmit = async function (event) {
   const browserSpecificFetch =
     navigator.userAgent.indexOf('Safari') + navigator.userAgent.indexOf('Chrome') < 0 ? content.fetch : window.fetch;
 
-  // load options
   const options = await getItemsFromStorage('options');
-  // generate the filename
   const pacerCaseId = event.data.id.match(/caseid=(\d*)/)[1];
   const filename = generateFileName(options, pacerCaseId);
 
--- a/src/content_delegate.js
+++ b/src/content_delegate.js
@@ -598,7 +598,7 @@ ContentDelegate.prototype.onDownloadAllSubmit = async function (event) {
   // load options
   const options = await getItemsFromStorage('options');
   // generate the filename
-  const pacerCaseId = event.data.id.match(/caseid\=\d*/)[0].replace(/caseid\=/, '');
+  const pacerCaseId = event.data.id.match(/caseid=(\d*)/)[1];
   const filename = generateFileName(options, pacerCaseId);
 
   // show loading message
--- a/src/content_delegate.js
+++ b/src/content_delegate.js
@@ -545,7 +545,16 @@ ContentDelegate.prototype.onDownloadAllSubmit = async function (event) {
     const page = document.createElement('html');
     page.innerHTML = html;
     const frames = page.querySelectorAll('iframe');
-    return frames[0].src;
+    if (frames.length) {
+      return frames[0].src;
+    }
+    // Try to extract the PDF URL from the HTML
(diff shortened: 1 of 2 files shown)


src/content_delegate.js
```diff
@@ -545,7 +545,16 @@ ContentDelegate.prototype.onDownloadAllSubmit = async function (event) {
     const page = document.createElement('html');
     page.innerHTML = html;
     const frames = page.querySelectorAll('iframe');
-    return frames[0].src;
+    if (frames.length) {
+      return frames[0].src;
+    }
+    // Try to extract the PDF URL from the HTML
+    const showTempURL = html.match(/\/cgi-bin\/show_temp.pl?(.*)/);
```
johnhawkinson, line 552: (While I'm otherwise here…

I am tempted to write this as

```
    const showTempURL = html.match(new Regexp('/cgi-bin/show_temp.pl?(.*)'));
```
to avoid escaping the forward slashes, but I'm not sure it's actually cleaner.
)

src/content_delegate.js
```diff
@@ -586,9 +595,34 @@ ContentDelegate.prototype.onDownloadAllSubmit = async function (event) {
   const browserSpecificFetch =
     navigator.userAgent.indexOf('Safari') + navigator.userAgent.indexOf('Chrome') < 0 ? content.fetch : window.fetch;
 
+  // load options
+  const options = await getItemsFromStorage('options');
+  // generate the filename
+  const pacerCaseId = event.data.id.match(/caseid\=\d*/)[0].replace(/caseid\=/, '');
```
johnhawkinson, line 601: OK, this is actually why I'm here. `\=` is not a valid escape sequence (outside of a v-mode character class) so this should just be `/caseid=\d*/`. I realize this problem was introduced before and you're just moving the code, but it should still be corrected.

src/content_delegate.js
```diff
@@ -586,9 +595,34 @@ ContentDelegate.prototype.onDownloadAllSubmit = async function (event) {
   const browserSpecificFetch =
     navigator.userAgent.indexOf('Safari') + navigator.userAgent.indexOf('Chrome') < 0 ? content.fetch : window.fetch;
 
+  // load options
+  const options = await getItemsFromStorage('options');
+  // generate the filename
```
johnhawkinson, line 600: (These are the kind of comments that convey no information.)

johnhawkinson (commented):
regexp caught my eye

src/content_delegate.js
ERosendo: You are correct! `\=` is not a valid escape sequence. Thank you for bringing this to our attention.

We can also use a capturing group and remove the replace() method.

src/content_delegate.js
ERosendo: 🤔 I used a regular expression literal because according to Eslint, it is considered the best practice.

This is the Eslint recommendation: 

> When a regular expression is known in advance, it is considered a best practice to avoid the string literal notation on top of the regular expression notation, and use regular expression literals instead of the constructor function.

You can read more about using this [link](https://eslint.org/docs/latest/rules/prefer-regex-literals#:~:text=When%20a%20regular%20expression%20is,instead%20of%20the%20constructor%20function.).

mlissner:
LGTM! Does this affect all zip uploads or just some of them for some reason?

ERosendo:
@mlissner I believe that all zip uploads made from a macOS device are affected. I tested the upload feature with different courts and got the same problem every time.

PR #350, opened 2023-11-17
feat(content_delegate): Use the URL constructor to return absolute URL

ERosendo:
During QA testing, I discovered that the zip upload feature was failing in **Firefox** and displaying an error message indicating that the extracted URL from the response was invalid. Further investigation revealed a known issue with Firefox's handling of relative URLs. To resolve this issue, we should employ absolute URLs instead of relative ones.

Here’s the Firefox issue [#2647](https://github.com/greasemonkey/greasemonkey/issues/2647).

Once this PR is merged, We'll be ready to release a new version containing the fix for ZIP uploads.

### DIFF
--- a/src/content_delegate.js
+++ b/src/content_delegate.js
@@ -554,7 +554,10 @@ ContentDelegate.prototype.onDownloadAllSubmit = async function (event) {
       return null;
     }
     // Clean the match found in the HTML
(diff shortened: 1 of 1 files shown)


mlissner (approved):

mlissner:
Good to see the QA script serving its purpose.

PR #373, opened 2024-04-30
feat(acms): Adds helper method to handle the Download Confirmation Page

ERosendo:
This PR adds support to upload documents from the new ACMS website and inserts a banner that lets users know a matching record is available in the RECAP archive.

The PR introduces the following changes:

- Refines the `dispatchHandler` method to effectively manage the **Download confirmation page**. The extension uses the `path` property for page identification.

- Uses the [MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) interface to monitor DOM changes and ensure all elements are rendered before proceeding. This PR removes the default `onclick` event of the **"Accept Charges"** button and replaces it with a custom event handler that allows us to upload documents to the RECAP archive.

- The Appellate class uses the ACMS API to retrieve the pdf documents so the PR adds a new file named `acms_api.js` to encapsulate utility functions designed to handle API interactions. I used the `exportInstance()` and `importInstance()` methods to enable the communication between the content script and the new service.

- Enhances the `parseReceiptPageTitle` method by refining the regular expressions to extract the data from the title of the download confirmation page. 

- Introduces a new helper function named `storeVueDataInSession` that extracts the app settings( API data and Auth token) and the data used by Vue to render the page and stores it in the `sessionStorage` object, allowing the extension access to the same data.

- Add a new parameter to the `uploadDocument` function so we can send the ACMS document GUI when it's available. 

Here's a gif showing how the extension works:

Screen Recording 2024-05-01 at 9 06 50 AM

and the banner looks like this: 

image

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,7 @@ The following changes are not yet released, but are code complete:
 Features:
  - Introduces a new upload type for Dockets from ACMS and enhances code maintainability by refactoring common upload logic into reusable functions.([#368](https://github.com/freelawproject/recap-chrome/pull/371))
  - Docket reports from ACMS are now uploaded to CourtListener.([#372](https://github.com/freelawproject/recap-chrome/pull/372))
+ - Adds logic to upload ACMS PDF documents to CourtListener. ([#373](https://github.com/freelawproject/recap-chrome/pull/373))
 
 Changes:
  - None yet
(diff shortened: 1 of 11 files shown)


mlissner (commented):
This seems fine to me, but it would be a lot better if you added more comments for each function. You're doing some clever things here that are in your brain now, but I think in a month or two you'll regret not explaining exactly what each thing does and why. 

Think you could add that, and then we can merge?

ERosendo:
> Think you could add that, and then we can merge?

Sure!

mlissner (approved):
Cool. Let's do it. Thank you!

PR #376, opened 2024-05-27
feat(acms): Adds logic to display the warning about combined PDFs

ERosendo:
This PR adds logic to check the number of documents on the download page and display the warning about combined PDFs when there's more than one.

Screen Recording 2024-05-27 at 5 58 33 PM


mlissner (approved):
LGTM. Let's see if I can resolve conflicts and then merge...

PR #378, opened 2024-07-05
feat(assets): Replace old logos with new ones

ERosendo:
Here are screnshots of the new logos and its variants:

- Recap in your Chrome extensions list(chrome://extensions/) page

image

- Inactive Recap Icon

image

- Active Recap Icon (Pacer Page)

image

- Recap Warning Icon (ACMS page): 

image

- Recap disabled upload:

image

- Banner about the pacer class action

image

- Banner about Recap Email

image

- Recap Icons and banner to create alert:

image

- Recap actions button:

image

- Recap Icon next to a docket entry:
image

- Button to accept charges and banner to let users know the document is available in the RECAP archive:
image

- Button and banner from the docket search page in PACER
image

- Warning about combined documents
image

- Options page: 
image


mlissner:
Looks ok, but the small icons need work. I think we should adjust the pixels by hand to balance the lines better. The middle one looks thick and I wouldn't want to look at it in my browser every day.

mlissner (commented):
Looks good, but let's refine these icons a bit more.

mlissner:
@ERosendo, I just pushed the latest icons as from Slack. If you're happy, let's merge. Before we ship, it'd be really nice to use our new notification/banner thing to let people know this change happened. Want to take a stab at that? I assume it's pretty easy, right?

ERosendo:
> Want to take a stab at that? 

Sure!

> I assume it's pretty easy, right?

Yes! @mlissner Would you like to display this banner just on the Options page, or consider including it on other PACER pages as well (similar to the login page banner)?

For the Options page: The banner will appear only when users click the recap icon, and it will look like this (I'll use a different message):

image

mlissner:
Yes, the way going forward is that all news goes on the login page and the options page, per the discussion here: https://github.com/freelawproject/recap/issues/367 (so doing this should fix that issue too).

ERosendo:
@mlissner How about adding a badge to the extension icon? This could be a great way to grab the attention of users who have pinned it to their browser toolbar. 🤔

mlissner:
Badge, login page, and options menu, please! Let's see if we can get people's attention.

src/assets/images/disabled-38.png
ERosendo, line 1: This icon should be gray, right?

ERosendo (commented):
Icons look great! Let's address the comment about the color of the `disabled` version and then merge.

mlissner:
Oops, I responded, but forgot to submit:

```
I thought the blue looked better and even makes more sense:

Gray vs. Blue: Are you on a PACER website

Slash vs. No Slash: Is the extension enabled
```

PR #379, opened 2024-07-12
feat(toolbar): Implements logic to display gift badge icon

ERosendo:
This pull request introduces several enhancements to the user interface:

- A new banner is added to the options page, complementing the existing banner on the login page.
- A badge is incorporated into the extension icon for improved visibility.

These changes provide greater flexibility and control for users:

- The close buttons on each banner operate independently, preventing accidental dismissals.
- The "Learn More" button offers a convenient way to remove both banners simultaneously.

Here are screenshots showing the banners:

image

image

Here's a gif showing how the "learn more" button works:

Screen Recording 2024-07-12 at 11 30 25 AM

This GIF demonstrates how users can hide a banner without losing access to the other:

Screen Recording 2024-07-12 at 11 43 15 AM


src/toolbar_button.js
```diff
@@ -28,6 +28,15 @@ function updateToolbarButton(tab) {
   };
 
   chrome.storage.local.get('options', function(items){
+    if (
+      'dismiss_news_badge' in items['options'] &&
+      items['options']['dismiss_news_badge']
+    ) {
+      chrome.browserAction.setBadgeText({ text: '' });
+    } else {
+      chrome.browserAction.setBadgeText({ text: '📣' });
```
mlissner, line 37: Everything looks good in this PR, but on my OS, the badge was a deep red color, which made it look like an error. I also wasn't quite able to figure out what the megaphone icon was without looking really closely. 

So two thoughts:

1. Can we change the background color to something that looks less like an error? Maybe blue? 

2. Can we use a more common emoji? I think the :gift: one is pretty common and what Slack uses?

mlissner (approved):
LGTM, but one usability thought below.

I think we'll continue refining this behavior (is it annoying to have to dismiss the news in two places?), but it's probably fine to roll with for now.

src/toolbar_button.js
ERosendo: > Can we change the background color to something that looks less like an error? Maybe blue?

We can customize the background color but I don't think we can tweak the style of the emoji. Here's an example:

image

image

> Can we use a more common emoji? 

I suggest using the bell emoji 🔔:  I just noticed the megaphone emoji renders as a red megaphone in Gmail, but the bell emoji 🔔 works fine.

image

mlissner: Great. Let's :shipit: to browsers!

ERosendo: 👍 I'll update the emoji

ERosendo:
@mlissner I've updated the emoji :shipit:

PR #385, opened 2024-07-21
fix(manifest): Refines content_scripts permissions on azure websites

ERosendo:
This PR updates the manifest file to implement a least privilege approach for permissions on Azure Websites. We're replacing the use of `include_globs` with a more granular approach that explicitly defines every single identified URL combination.

This change addresses concerns about overly broad permissions previously granted by the `matches` property. 

Fixes https://github.com/freelawproject/recap/issues/374

Here are screenshots showcasing the revised permission list. Notably, the option granting access to all Azure Websites sites is no longer included.

image

image

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,7 +11,7 @@ Changes:
  - None yet
  
 Fixes:
- - None yet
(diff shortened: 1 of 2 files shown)


mlissner (approved):

mlissner:
Thank you

PR #388, opened 2024-08-27
feat(build): Adds automated release package creation script

ERosendo:
This pull request introduces a new script, `build.sh`, to address inconsistencies in how **Chrome** and **Firefox** handle the `favicon_url` key within the `search_provider` object. This script automates the creation of release packages specific to each browser.

Key changes: 

- Ensures consistent behavior across **Chrome** and **Firefox** by addressing the `favicon_url` inconsistency.

- Introduces two new commands in `package.json`:
   - `npm run release-chrome`: Creates a Chrome-compatible release package within a dedicated build folder. 
   - `npm run release-firefox`: Creates a Firefox-compatible release package within the same build folder.

- Removes the manual creation of `zip` files, streamlining the release process.

- Updates the `README.md` file to reflect new build steps


scripts/build.sh
```diff
@@ -0,0 +1,47 @@
+# Create a browser extension release package
```
mlissner, line 1: You'll want to add a shebang here, and adding `set -e` makes it crash completely if there's an error on any line (instead of idiotically continuing like a chicken with its head cut off):

```
#!/bin/bash
set -e
```

scripts/build.sh
```diff
@@ -0,0 +1,47 @@
...
+browserType=$1
+
+# 2. Construct release zip filename
+zipName=$browserType-release.zip
+
+# 3. Ensure release directory exists (create if needed)
+mkdir -p build/release/
+
+# 4. Clean up any existing release zip
+rm -f build/release/$browserType-release.zip
```
mlissner, line 13: Do you need to force this? Scary stuff.

scripts/build.sh
```diff
@@ -0,0 +1,47 @@
...
+
+# 5. Rename base manifest file
+cd src/ && mv manifest.json manifest.base.json
+if [[ "$browserType" == "firefox" ]]; then
+    faviconUrl='assets/images/favicon.icon'
+    # Remove unnecessary properties for Firefox manifest
+    jq 'del(.background.service_worker, .background.type)' manifest.base.json >manifest.json
```
mlissner, line 20: We should do a check at the beginning and throw an error if `jq` is missing, before we do other stuff.

scripts/build.sh
mlissner, line 13: I think you can move steps 3 and 4 down below where you need them, right? No need to do them here?

scripts/build.sh
```diff
@@ -0,0 +1,47 @@
...
+# - Include all files except the base manifest
+zip -rq $zipName * -x "*.base.json"
+
+# 8. Move package to release directory
+mv $zipName ../build/release
+
+# 9. Remove Browser-Specific Manifest File
+rm manifest.json
+# Rename base manifest back to main manifest
+mv manifest.base.json manifest.json
```
mlissner, line 47: Pretty sure `mv` is a bastard that doesn't mess around and you can just do:

```suggestion
# 9. Overwrite browser-specific file with original
mv manifest.base.json manifest.json
```

mlissner (approved):
Looks good. I fear and loathe bash, so I wound up insulting it a bit and adding a number of comments. Easy stuff though. LGTM once you're ready.

scripts/build.sh
ERosendo: You're right!

scripts/build.sh
ERosendo: that's a good idea

scripts/build.sh
ERosendo: > I think you can move steps 3 and 4 down below where you need them, right?

You're totally right!

scripts/build.sh
ERosendo: 😮 😮

scripts/build.sh
ERosendo: > Do you need to force this? 

Not really, I used the `-f` flag to suppress the exit code when the file couldn't be found. However, I've modified the command to handle missing files differently. Thanks for your comment.

mlissner (approved):

PR #391, opened 2024-08-30
fix(safari): Resolve manifest upgrade compatibility issues

ERosendo:
This PR addresses the following issues: 

- **Updates the script that creates the MacOS and iOS version of the extension**: After updating the manifest file to include new API permissions, the `update_manifest_files` helper method was overwriting the existing permissions for the macOS and iOS versions of the extension with a hardcoded list(This caused compatibility issues). The PR removes the hardcoded permissions list from the helper method, ensuring that the same set of permissions is used across all browsers.
 
- While testing the extension, I noticed that the code that creates notifications was consistently raising exceptions.  Investigating the [Notifications API documentation ](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/notifications/create), I realized **Safari** currently doesn't support this functionality. To prevent these errors and ensure a smooth user experience across browsers, I implemented an early return mechanism within the `showNotification` function. This return now occurs specifically when the extension detects it's running in **Safari**.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,10 +11,10 @@ Changes:
  - None yet
  
 Fixes:
- - None yet
+ - Disables notifications in Safari due to API limitations ([391](https://github.com/freelawproject/recap-chrome/pull/391)).
 
 For developers:
- - Nothing yet
+ - Preserves permissions during macOS and iOS release package creation ([391](https://github.com/freelawproject/recap-chrome/pull/391))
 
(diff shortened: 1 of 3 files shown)


mlissner:
Ugh. We used to ask people to install a special notification tool on Apple computers to do notifications, because the OS hadn't figured them out yet. 

I guess a solution here could be to use the website's notification system, assuming Safari supports that. Seems like it can wait, but I guess it's a solution.

PR #394, opened 2024-09-11
feat(toolbar): Enhance banners with updated content

ERosendo:
This PR removes the banner from the login page and updates the banner within the options page to inform users about the code upgrade to Manifest Version 3. Additionally, it updates the badge icon style to use a number instead of the bell emoji for Firefox and Chrome.

Here are screenshots showing the banner and the badge icon:

image

-----

image

----

Screen Recording 2024-09-11 at 5 07 20 PM

### DIFF
--- a/src/content.js
+++ b/src/content.js
@@ -47,31 +47,6 @@ async function addRecapInformation(msg) {
       });
     }
 
-    const options = await getItemsFromStorage('options');
-    if (
-      'login_dismiss_new_brand_info' in options &&
-      options['login_dismiss_new_brand_info']
-    ) {
-      return;
-    }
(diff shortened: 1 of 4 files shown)


mlissner (commented):
Looks good, but don't we need to reset the cookie or something so that it shows up when the new version is installed?

ERosendo:
> don't we need to reset the cookie or something so that it shows up when the new version is installed?

@mlissner You're right! I've addressed this issue by replacing the existing flag in `localStorage` with a new one that ensures the banner is displayed upon installation.

mlissner (approved):

PR #398, opened 2024-09-18
fix(district): Attaches pacer_case_id to entry links

ERosendo:
This PR introduces logic to attach the `pacer_case_id` as a query parameter to entry links. This updates ensures the case ID is  easily accessible from the URL even if it's not present in the subsequent page's DOM. This will simplify the process of navigating between pages while retaining important case information.

This PR partially fixes https://github.com/freelawproject/recap/issues/369. Users accessing attachment pages from CourtListener or the docket report will no longer encounter problems. However, the extension's ability to capture attachment pages remains limited when users employ links from sources outside these two.

@mlissner If the extension is unable to gather all the necessary data for a page upload, should we display a warning message to alert users of the potential issue and guide them towards a successful upload?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,7 @@ Changes:
  
 Fixes:
  - Corrected typo in build script, ensuring correct favicon path for Firefox releases([379](https://github.com/freelawproject/recap/issues/379), [397](https://github.com/freelawproject/recap-chrome/pull/397))
+ - Attaches the pacer_case_id as a query parameter to entry links in District Courts Reports ([369](https://github.com/freelawproject/recap/issues/369), [398](https://github.com/freelawproject/recap-chrome/pull/398)).
 
 For developers:
  - Nothing yet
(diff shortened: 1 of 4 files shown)


mlissner:
Hm, I think adding things to the URL should be the last resort since such parameters can get sent to the server and since they impact the page itself (via things like `replaceNode`, for example).

Looking at the attachment page in the issue: 

https://ecf.cacd.uscourts.gov/doc1/031038943352?caseid=868080

The HTML seems to have the case ID in the goDLS thing:

```html
<a href="https://ecf.cacd.uscourts.gov/doc1/031138943352" onclick="goDLS('/doc1/031138943352','868080','9','','','','','','');return(false);">1</a>
```

Why not parse it from there? We've got utilities for that, right?

ERosendo:
> Why not parse it from there? We've got utilities for that, right?

When attempting to parse the goDLS function using the existing method, I encountered an issue. The current implementation relies on a regular expression designed to extract eight parameters. However, this version of goDLS uses nine parameters.

I'll update the helper method and adjust the overall approach of this PR to retrieve the `pacer_doc_id` from the goDLS function.

Thanks for your comment.

mlissner:
Interesting. Now we can try to figure out what that ninth one is!

PR #403, opened 2024-10-03
Fix setDefaultOptions in updateToolbarButton

ttys0dev:
This check needs to be before items is accessed.


ERosendo:
Thanks @ttys0dev, LGTM :shipit:

PR #404, opened 2024-10-04
fix(background): use chrome.scripting.ExecutionWorld.MAIN for firefox compatibility

ttys0dev:
It seems firefox is picky here and won't accept the a `MAIN` string.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,6 +12,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         node-version: [14.x, 16.x, 18.x]
     steps:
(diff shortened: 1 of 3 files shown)


ttys0dev:
rebased

ttys0dev:
@ERosendo Does this look good to merge?

mlissner:
@ttys0dev, what's the impact of this being wrong?

ttys0dev:
> what's the impact of this being wrong?

I was seeing an exception being thrown.

mlissner:
What version of FF?

ttys0dev:
> What version of FF?

Was some ESR version I think where I saw the error, may have updated it since I tested though(currently on version 128 looks like).

PR #405, opened 2024-10-05
fix(notifier): call sendResponse when notifications are disabled

ttys0dev:
The callers expect a response even if we don't actually show a notification.


ERosendo:
Thanks @ttys0dev

PR #406, opened 2024-11-16
Handle attachment number rows with six child elements

ttys0dev:
Looks like some of these(like ca1) have six elements.


mlissner:
Looks like we just need some release notes and we can merge.

ERosendo:
@mlissner I'll handle it.

PR #407, opened 2024-11-25
fix(appellate): Avoids sending attachment number 0

ERosendo:
While assisting with the debugging of https://github.com/freelawproject/courtlistener/issues/3982, I noticed that a significant number of extension uploads originated from appellate courts. Upon investigating the underlying appellate helper functions, I discovered that some were using 0 as a default value, which leads to unexpected behavior.

This PR addresses this issue by updating the helper functions to use `null` as the default value instead of `0`.

Fixes https://github.com/freelawproject/recap/issues/382


mlissner (approved):
Great!

PR #411, opened 2024-12-30
fix(utils): Improve Banner Insertion Reliability for regular users

ERosendo:
This PR improves the reliability of banner insertion within the `checkSingleDocInCombinedPDFPage` function. The previous target element for **district courts**, now has the hidden attribute, which could interfere with banner display. This change addresses potential inconsistencies by switching the target element to the main div.


elisa-a-v (approved):
LGTM!

PR #412, opened 2025-01-21
feat(appellate): Adds guarding logic for missing pacer_doc_id

ERosendo:
This PR adds a check to ensure that the `pacer_doc_id` is available before attempting to retrieve the PDF. If the `pacer_doc_id` is missing, the code will exit early to prevent unintended modifications to the page and incorrect `GET` requests.
Screen Recording 2025-01-21 at 8 59 35 PM

This PR also updates the `karma.config.js` file to fix some failing tests 

Fixes https://github.com/freelawproject/recap/issues/388

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,7 +5,7 @@
 The following changes are not yet released, but are code complete:
 
 Features:
- - None yet
+ - More robust guarding logic for missing `pacer_doc_id` in appellate courts ([388](https://github.com/freelawproject/recap/issues/388), [412](https://github.com/freelawproject/recap-chrome/pull/412))
 
 Changes:
  - None yet
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -29,7 +29,8 @@ module.exports = function(config) {
       'src/utils/fetch.js',
       'src/utils/notifier.js',
       'src/pacer.js',
-      'src/toolbar_button.js',
+      { pattern: 'src/utils/url_and_cookie_helpers.js', type: 'module' },
+      { pattern: 'src/utils/toolbar_button.js', type: 'module' },
       'src/utils.js',
       'src/action_button.js',
       'src/pdf_upload.js',
@@ -73,7 +74,7 @@ module.exports = function(config) {
 
     // start these browsers
     // available browser launchers: https://npmjs.org/browse/keyword/karma-launcher
-    browsers: ['ChromeHeadless'],
+    browsers: ['CustomChromeHeadless'],
 
     // set these options to view logs in development
     // see https://github.com/karma-runner/karma/issues/2582#issuecomment-413660796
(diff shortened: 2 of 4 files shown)


mlissner (approved):
Seems simple enough, and LGTM. Let's see what Elisa thinks!

elisa-a-v (changes requested):
When following the steps described [here](https://github.com/freelawproject/recap/issues/388#issuecomment-2605925999) to get docs from [this docket](https://www.courtlistener.com/docket/69552415/joshua-jordan-v-payment-saver-llc/), I got this page:
image

and the doc button brought me here:
image

I then selected the second attachment and the URL didn't include the doc ID (`https://ecf.ca4.uscourts.gov/docs1/004110098892?caseId=null&recapAttNum=2`), but the GET request did include it in the query params so that works!

~Sadly, I noticed despite getting a 200 status response, the `results` list was empty, even though we have the docket, the docket entry and the RECAPDocument for that attachment (as seen in CL).~ Well, upon closer inspection I realized this part is actually fine, since we don't have the PDF for that RECAPDocument.

image
image

I then bought the document and could see it in the browser, but the POST request failed with a 400 status, because there was no `pacer_doc_id` in the payload:
image

I did not try to reproduce the issue with the live extension so as not to add to the DB issue 😅 so I'm not entirely sure this is a consequence of this PR or not, what do you think @ERosendo ?

ERosendo:
Hey @elisa-a-v Thanks for your feedback! You have a great eye for catching cases that can break the code. 

Your detailed report helped me identify an issue where the extension was incorrectly adding `caseId=null` to the attachment URLs.

Commit [a538050](https://github.com/freelawproject/recap-chrome/pull/412/commits/a5380503391382d10a665652c2670232d9547f30) resolves this. I've tested the extension with the specific case you mentioned in your comment, and it successfully retrieved the document as expected.

Screen Recording 2025-01-24 at 12 20 16 PM

ERosendo:
I also believe adding support for this new page in the future would be beneficial. However, it falls outside the scope of this current PR, which is focused on fixing the `GET` request triggering the db performance issue.

ERosendo:
@elisa-a-v Thanks for the steps you outlined in your most recent comment! I followed them, but it seems we don't have enough data to upload the file on the document purchase page.

To address this, I made a small adjustment to the logic we introduced in  https://github.com/freelawproject/recap-chrome/pull/402. This change will make the extension return early in cases where the document ID can be retrieved. This way, we avoid adding unnecessary buttons or elements that might confuse users into thinking the extension is trying to upload the document.

I reviewed the `GET` requests and confirmed that the extension correctly sends the `doc_id` and avoids triggering the same problematic database query.

I think the problem with the combined PDF page should be included in issue #389 (https://github.com/freelawproject/recap/issues/389), which aims to add support for the Orders/Judgments section.

elisa-a-v (approved):
Nice, I think the early return is the right call here.

I say we should merge this and dig deeper into the other stuff whenever we get to https://github.com/freelawproject/recap/issues/389

PR #413, opened 2025-02-26
feat(district): Clears docketNumber from localStorage

ERosendo:
This PR fixes  https://github.com/freelawproject/recap/issues/392: `docketNumber` values causing errors after viewing district court reports. Unlike appellate court reports, district court reports lack a consistent source for docket numbers, making extraction unreliable. Therefore, this commit clears the `docketNumber` from **localStorage** when a district court report is loaded, ensuring clean state and preventing unintended use of outdated data.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,7 +11,7 @@ Changes:
  - None yet
 
 Fixes:
- - None yet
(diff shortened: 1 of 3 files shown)


mlissner (commented):
Huh. Seems like a simple enough fix. Let's have Elisa give it a second look too.

elisa-a-v (approved):
LGTM! 🚀

PR #415, opened 2025-06-10
feat(options): Removes banner about manifest v3

ERosendo:
Key changes:

- The banner in `options.html` is now hidden, and the header height has been adjusted. This ensures proper alignment of buttons and menu content, creating a clean layout.

- The default settings have been updated to hide the news badge on fresh installations.

fixes https://github.com/freelawproject/recap/issues/395

### DIFF
--- a/src/options.html
+++ b/src/options.html
@@ -9,11 +9,11 @@
 </head>
 
 <body id="options-body">
-  <div id="container" class="grid-with-banner">
+  <div id="container" class="regular-grid">
     <!-- Main Tab Content + Donation plea -->
     <header>
       <div class="d-flex flex-column">
-        <div id="header-banner" class="info-banner d-flex flex-column">
+        <div id="header-banner" hidden class="info-banner d-flex flex-column">
           <div id="message-section" class="d-flex">
             <div class="flex align-items-center banner-icon-container">
               <div id="banner-icon">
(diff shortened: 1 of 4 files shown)


src/options.html
```diff
@@ -9,11 +9,11 @@
 </head>
 
 <body id="options-body">
-  <div id="container" class="grid-with-banner">
+  <div id="container" class="regular-grid">
     <!-- Main Tab Content + Donation plea -->
     <header>
       <div class="d-flex flex-column">
-        <div id="header-banner" class="info-banner d-flex flex-column">
+        <div id="header-banner" hidden class="info-banner d-flex flex-column">
```
mlissner, line 16: This can't be right, can it? 

Perhaps you meant:

```suggestion
        <div id="header-banner" class="info-banner d-flex flex-column hidden">
```

?

mlissner (changes requested):

src/options.html
ERosendo: The extension uses Bootstrap v4.3.1, and the `hidden` attribute seems to be fully supported:

https://getbootstrap.com/docs/4.1/content/reboot/#html5-hidden-attribute

mlissner: OMG, I had no idea. They did it, they really did.

mlissner (approved):

PR #425, opened 2026-03-05
chore: update dependencies

dalelotts:
Please squash merge only

PR #426, opened 2026-03-12
chore: update dependencies

dalelotts:
Please squash merge only

Supersedes #425.

Superseded by #427.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,11 +11,16 @@ Changes:
  - None yet
 
 Fixes:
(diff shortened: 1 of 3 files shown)


mlissner:
Hi Dale, thanks for all this work! A few things at a glance: 

1. I worry about the risk of accepting so many packages like this when we can't review the vendor'ed code for possible vuln injection. I wonder if it would be smart to move some of these to something like a package lock file so that they can come from there in a build script, instead of just being packaged like this? 

2. This is really big. Whether or not we approach it as in 1, above, I wonder if splitting it apart makes sense so that we can just upgrade a few things at a time? 

3. Aside from the above, there are some unrelated files in here. Can you remove them, please? 

Thank you!

dalelotts:
Whoa! 😱 I intended to just restore the previous PR after deleting my fork. 

Yeah, I'll get rid of all of this junk and get back to only what #425 contained.

PR #427, opened 2026-03-15
build: update dependencies

dalelotts:
Supersedes #425 and #426 

Fix the required checks:

- go to repo `Settings`
- `Branches` or `Rulesets`
- find the protection/ruleset for `main`
- remove required checks:
  - `build (14.x)`
  - `build (16.x)`
  - `build (18.x)`
- add:
  - `build (20)`
  - `build (22)`
  - `build (24)`
  
  
Please squash merge only.


dalelotts:
Net -23K lines. Not too bad. =)

dalelotts:
If this is still too much I'm happy to trim it down further.

mlissner:
Thanks, this is helpful. I'm still worried about the JS libraries that we vendor here. I wonder if you could include a script for us to download those instead of including them directly? You seem like a real and normal person, but I'm thinking more about the general case of this kind of thing, because we get a lot of contributions and I know extensions like ours are a vector for clever attacks. 

Would that be feasible as a way of making this PR smaller and safer?

dalelotts:
Yeah, no problem. It's fine, you don't know me, you should be cautious. (if you did know me, you might be more cautious LOL). I appreciate the cautious approach and will not take it personally. 

I didn't bother reading any contributing documentation, I'll admit I'm lazy =). I did a quick scan and didn't see any guidance for updating dependencies, etc.  Is there some guidelines I can read?

> I'm still worried about the JS libraries that we vendor here.

I'm not sure what you mean here, can you elaborate?  I did not introduce any new libraries, just new versions, and I assume FreeLaw has a vetted binary repository to resolve dependencies and an audit process for adding new versions (both of which I assumed are out-of-scope for any PR). 

I deleted any checked in copies of libraries (from assets/js) thinking that was okay, because package.json versions are NOT locked - (nor were they locked in CI - using npm install over npm ci) - so it seems to me that vendor-copy could inject newer versions of those scripts as they became available. (maybe you don't want to use `^` or `~` in package.json)

I think Intl.RelativeTimeFormat can be trusted given it is builtin - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat


Happy to adjust as you see fit, I just need a bit of direction from you about what FreeLaw would consider a safer PR.

mlissner:
> I did a quick scan and did't see any guidance for updating dependencies, etc. Is there some guidelines I can read?

No, no guidance yet. 

> I'm not sure what you mean here, can you elaborate? I did not introduce any new libraries, just new versions, and I assume FreeLaw has a vetted binary repository to resolve dependencies and an audit process for adding new versions (both of which I assumed are out-of-scope for any PR).

I meant new versions. We don't have a process for this yet. We're making it right now. We need a way of being sure the code we're landing is valid. The best way I can think of to do that is for you to provide a script with links to versions we need to download. Or NPM, if it can do what we want.

Will that work? If not, I'll hand this off to a dev who will have more time to think about it more carefully (hopefully!)

dalelotts:
yeah, whatever you need. The versions are in package.json so npm is the typical path to get those. npm will pull in all of the transitive dependencies so it will be more than the deltas in package.json 

Let me see if I can get a report or something from npm on the product. 

I see I downgraded bootstrap - prod already has 5, but I downgraded because most of the html I saw was still using markup from 4. I will go back to 5 and assume the 4 markup is dead.

Once I switch back to bootstrap 5, then the only change to what goes out to the user is jquery and web-ext - the dev dependencies are not distributed (at least they should not be, I did not audit the build to make sure)

I'll work on some changes. Hold please.

dalelotts:
Bootstrap changed from ^5.2.3 to ^5.3.8 but both of those specs will install 5.3.8 so no issue.

JQuery changed from ~3.6.3 to ~3.7.1 so still v 3 but not exactly the same version - do you have other tools already using ~3.7.1? if so, can I assume it is already vetted?

web-ext changed from ^7.5.0 to ^10.0.0 so that's the biggest version jump. Again, is anything else using ^10.0.0?

The rest are dev dependencies and should not be shipped to prod.

This is what I found so far (obviously someone in FreeLaw should run these audits themselves)

```
npm audit
found 0 vulnerabilities
```

```
npm audit signatures                               
audited 647 packages in 2s

647 packages have verified registry signatures

51 packages have verified attestations
```
A json report of the audit generated via `npm audit --json`
[recap-audit.json](https://github.com/user-attachments/files/26039269/recap-audit.json)

A software bill of materials generated via `npm sbom --sbom-format cyclonedx`
[recap-sbom.json](https://github.com/user-attachments/files/26039270/recap-sbom.json)

These appear to include transitive dependencies for prod and dev.

mlissner:
@ERosendo, to you for prioritization. I think this just needs a review, but I want to be sure that any vendored code is authentic. Let's assume Dale is a bad guy and that he sneaked a terrible virus into the minified code. 

Thank you both!

PR #428, opened 2026-07-21
fix(district): stop attributing docket uploads to a stale cached case id

playfulart:
## What this fixes

When a district docket report page yields no case id from its own context (URL, inputs, referrer), `handleDocketDisplayPage` falls back to the **per-tab cached case id** and uploads the entire page HTML under it. That cached value can belong to a **different case viewed earlier in the same tab** — nothing invalidates it on navigation — so one case's full docket sheet gets uploaded under another case's `pacer_case_id`, and the archive faithfully merges one case's entire history (entries *and* parties) into the other's docket.

## Real-world occurrence

This happened to a pro se litigant's case in D. Nev.: the complete docket of *Empire Trust v. Cellura* (nvd `3:25-cv-00553`, pacer_case_id `177277`) was uploaded under *Mitchell v. Tate*'s id (nvd `2:26-cv-00720`, pacer_case_id `180126`). CourtListener docket `72456248` now contains ~165 entries byte-identical (same entry numbers, dates, text, and `pacer_doc_id`s) to entries on docket `71566874`, including the foreign case's SDNY pre-transfer history and two of its parties. Both CL docket records carry correct, distinct `pacer_case_id`s, so the wrong id had to be attached client-side at upload time. A data-repair request for that docket was sent separately via the CourtListener contact form.

## The fix

Before reaching for the cached id, derive the case id from the page itself: a district docket sheet carries its own case id in every document link's `goDLS()` onclick handler. New helper `PACER.getCaseIdFromDocketDisplayLinks()` tallies those ids, and the fallback path in `handleDocketDisplayPage` now prefers that evidence over the tab cache:

- The **plurality** (most common) id wins, so a stray link to another case's document — legitimate on consolidated/MDL sheets (the existing comment in `findAndStorePacerDocIds` notes this corner case) — cannot outvote the sheet itself.
- **A tie counts as no evidence** — the cached-id fallback applies as before rather than guessing.
- **An id the page context already provided is never overridden.** On consolidated *member* dockets the document links legitimately point at the *lead* case, so the goDLS plurality is only trusted when there is no context id at all. This keeps the change strictly scoped to the previously-unprotected fallback path — behavior on every other path is byte-for-byte unchanged.
- When the page id beats a conflicting cached id, a console warning says so, and the cached id is **corrected**, seeding the right id for a later docket-report load in the same tab.
- **Pages with no goDLS evidence behave exactly as before** (some courts don't emit goDLS).

## Tests

Five new specs in `ContentDelegateSpec.js`:

- stale tab-cached id + page links assert another id → upload uses the page's id, nothing is sent under the stale id, and the cached id is corrected in storage;
- consolidated **member** docket shape: context-derived id present + goDLS plurality pointing at the lead case → context id wins (never overridden);
- consolidated plurality: three links to the lead case + one stray member-case link → most common id used;
- tied ids → no guessing, cached-id fallback preserved;
- fresh tab with no storage record at all → no throw, no upload (regression for the missing-record guard caught in review).

Full suite: **188/188 passing** (183 pre-existing + 5 new). `CHANGES.md` updated under Upcoming → Fixes.

## Review round 1 (@ERosendo)

All four suggestions adopted, one commit each: the sort-based rewrite of `getCaseIdFromDocketDisplayLinks`, the plurality (not strict-majority) doc wording, the corrected comment on what the tab-storage write feeds (a later docket-report load — attachment menus resolve via `docsToCases`, verified), and the `|| {}` guard on `getItemsFromStorage` with the regression spec above.

### DIFF
--- a/src/content_delegate.js
+++ b/src/content_delegate.js
@@ -286,8 +286,8 @@ ContentDelegate.prototype.handleDocketDisplayPage = async function () {
         );
       }
       this.pacer_case_id = pageCaseId;
-      // Correct the cached id so caseless pages downstream (e.g.
-      // attachment menus) inherit the right one.
+      // Seed the tab's cached case ID so a later docket-report load in this
+      // same tab starts with the correct case ID instead of a stale one.
       await saveCaseIdinTabStorage({ tabId: this.tabId }, pageCaseId);
     } else {
       this.pacer_case_id = tabStorage.caseId;
--- a/src/content_delegate.js
+++ b/src/content_delegate.js
@@ -269,12 +269,30 @@ ContentDelegate.prototype.handleDocketDisplayPage = async function () {
   // check if appellate
   // let isAppellate = PACER.isAppellateCourt(this.court);
 
-  // if the content_delegate didn't pull the case Id on initialization,
-  // check the page for a lead case dktrpt url.
-  const tabStorage = await getItemsFromStorage(this.tabId);
-  this.pacer_case_id = this.pacer_case_id
-    ? this.pacer_case_id
-    : tabStorage.caseId;
+  // If initialization yielded no case id, derive it from the page itself:
+  // the sheet's goDLS document links carry it. The per-tab cached id is
+  // the LAST resort — it can belong to a different case viewed earlier in
+  // this tab. An id the page context already provided is never overridden:
+  // consolidated member dockets legitimately link the lead case's
+  // documents, so the goDLS majority only speaks when nothing else does.
+  if (!this.pacer_case_id) {
+    const pageCaseId = PACER.getCaseIdFromDocketDisplayLinks(this.links);
+    const tabStorage = await getItemsFromStorage(this.tabId);
+    if (pageCaseId) {
+      if (tabStorage.caseId && tabStorage.caseId !== pageCaseId) {
+        console.warn(
+          `RECAP: Ignoring cached case id ${tabStorage.caseId}: the ` +
+            `docket's own links say it belongs to case ${pageCaseId}.`
+        );
+      }
+      this.pacer_case_id = pageCaseId;
+      // Correct the cached id so caseless pages downstream (e.g.
+      // attachment menus) inherit the right one.
+      await saveCaseIdinTabStorage({ tabId: this.tabId }, pageCaseId);
+    } else {
+      this.pacer_case_id = tabStorage.caseId;
+    }
+  }
 
   // If we don't have this.pacer_case_id at this point, punt.
   if (!this.pacer_case_id) return;
--- a/src/content_delegate.js
+++ b/src/content_delegate.js
@@ -274,7 +274,7 @@ ContentDelegate.prototype.handleDocketDisplayPage = async function () {
   // the LAST resort — it can belong to a different case viewed earlier in
   // this tab. An id the page context already provided is never overridden:
   // consolidated member dockets legitimately link the lead case's
-  // documents, so the goDLS majority only speaks when nothing else does.
+  // documents, so the goDLS plurality only speaks when nothing else does.
   if (!this.pacer_case_id) {
     const pageCaseId = PACER.getCaseIdFromDocketDisplayLinks(this.links);
     const tabStorage = await getItemsFromStorage(this.tabId);
--- a/src/content_delegate.js
+++ b/src/content_delegate.js
@@ -277,7 +277,7 @@ ContentDelegate.prototype.handleDocketDisplayPage = async function () {
   // documents, so the goDLS plurality only speaks when nothing else does.
   if (!this.pacer_case_id) {
     const pageCaseId = PACER.getCaseIdFromDocketDisplayLinks(this.links);
-    const tabStorage = await getItemsFromStorage(this.tabId);
+    const tabStorage = (await getItemsFromStorage(this.tabId)) || {};
     if (pageCaseId) {
       if (tabStorage.caseId && tabStorage.caseId !== pageCaseId) {
         console.warn(
--- a/src/pacer.js
+++ b/src/pacer.js
@@ -462,11 +462,11 @@ let PACER = {
     }
   },
 
-  // Returns the case id a docket report page asserts about itself: the
-  // strict majority of the de_caseid values in its document links' goDLS()
-  // handlers. Majority, because consolidated/MDL sheets legitimately link
-  // member cases' documents. Returns undefined when there is no goDLS
-  // evidence or the top ids are tied.
+  // Returns the most common (plurality) `de_caseid` across the page's
+  // document links' `goDLS()` handlers. Plurality is intentional because
+  // consolidated/MDL docket sheets can legitimately link documents from
+  // multiple member cases. Returns `undefined` when there is no `goDLS`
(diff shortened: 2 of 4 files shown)


src/pacer.js
```diff
@@ -462,6 +462,35 @@ let PACER = {
     }
   },
 
+  // Returns the case id a docket report page asserts about itself: the
+  // strict majority of the de_caseid values in its document links' goDLS()
+  // handlers. Majority, because consolidated/MDL sheets legitimately link
+  // member cases' documents. Returns undefined when there is no goDLS
+  // evidence or the top ids are tied.
+  getCaseIdFromDocketDisplayLinks: function (links) {
```
ERosendo, line 470: I wonder if we could simplify this function a bit to make it easier to follow.

The current tally-then-tie logic took me a second to parse.

What do you think about using more descriptive variable names and sorting the counts before checking for a tie? I found that it makes the algorithm read more naturally from top to bottom, while keeping the same behavior. It also lets the comments describe the intent a bit more clearly.


```js
  getCaseIdFromDocketDisplayLinks: function (links) {
    // Count how many document links point to each case ID. The case ID
    // appearing most often is likely the case represented by the docket page.
    const caseIdCounts = {};
    for (const link of Array.from(links)) {
      if (!PACER.isDoc1Url(link.href)) continue;

      const goDLS = PACER.parseGoDLSFunction(link.getAttribute('onclick'));
      if (goDLS && goDLS.de_caseid && goDLS.de_caseid !== '0') {
        caseIdCounts[goDLS.de_caseid] =
          (caseIdCounts[goDLS.de_caseid] ?? 0) + 1;
      }
    }

    // Convert the counts object into an array of [caseId, count] pairs
    // and sort descending so the most frequently linked case is first.
    const sortedCaseIds = Object.entries(caseIdCounts).sort(
      (a, b) => b[1] - a[1]
    );

    // No valid case IDs were found.
    if (sortedCaseIds.length === 0) return undefined;

    const [mostLinkedCaseId, mostLinkedCount] = sortedCaseIds[0];
    const [, secondMostLinkedCount] = sortedCaseIds[1] ?? [];

    // If multiple case IDs have the same highest count, we cannot determine
    // which case is the correct one, so avoid returning an arbitrary result.
    const hasTieForFirst = mostLinkedCount === secondMostLinkedCount;

    return hasTieForFirst ? undefined : mostLinkedCaseId;
  },
```

src/pacer.js
```diff
@@ -462,6 +462,35 @@ let PACER = {
     }
   },
 
+  // Returns the case id a docket report page asserts about itself: the
+  // strict majority of the de_caseid values in its document links' goDLS()
+  // handlers. Majority, because consolidated/MDL sheets legitimately link
```
ERosendo: I think "strict majority" is a little misleading here. The function is actually looking for the plurality (the most common case ID), not a value that's present in more than 50% of the links. For example, `{A: 3, B: 2, C: 2}` correctly returns A.

The current behavior seems right for our use case. I'm only suggesting updating the wording so it matches what the code does. 

```js
// Returns the most common (plurality) `de_caseid` across the page's 
// document links' `goDLS()` handlers. Plurality is intentional because 
// consolidated/MDL docket sheets can legitimately link documents from 
// multiple member cases. Returns `undefined` when there is no `goDLS` 
// evidence or when the most common case IDs are tied.
```

src/content_delegate.js
```diff
@@ -269,12 +269,30 @@ ContentDelegate.prototype.handleDocketDisplayPage = async function () {
...
+    if (pageCaseId) {
+      if (tabStorage.caseId && tabStorage.caseId !== pageCaseId) {
+        console.warn(
+          `RECAP: Ignoring cached case id ${tabStorage.caseId}: the ` +
+            `docket's own links say it belongs to case ${pageCaseId}.`
+        );
+      }
+      this.pacer_case_id = pageCaseId;
+      // Correct the cached id so caseless pages downstream (e.g.
+      // attachment menus) inherit the right one.
```
ERosendo: I think this comment could be a little more precise about what this write does.

the attachment menus read the `docsToCases` map (`getPacerCaseIdFromPacerDocId`), not the `caseId` we're updating here, so this value isn't used by them directly. What this does do (which is still useful!) is seed the correct case ID for a later docket-report load in the same tab.

```js
// Seed the tab's cached case ID so a later docket-report load in this 
// same tab starts with the correct case ID instead of a stale one. 
await saveCaseIdinTabStorage({ tabId: this.tabId }, pageCaseId);
```

src/content_delegate.js
```diff
@@ -269,12 +269,30 @@ ContentDelegate.prototype.handleDocketDisplayPage = async function () {
...
-    : tabStorage.caseId;
+  // If initialization yielded no case id, derive it from the page itself:
+  // the sheet's goDLS document links carry it. The per-tab cached id is
+  // the LAST resort — it can belong to a different case viewed earlier in
+  // this tab. An id the page context already provided is never overridden:
+  // consolidated member dockets legitimately link the lead case's
+  // documents, so the goDLS majority only speaks when nothing else does.
+  if (!this.pacer_case_id) {
+    const pageCaseId = PACER.getCaseIdFromDocketDisplayLinks(this.links);
+    const tabStorage = await getItemsFromStorage(this.tabId);
```
ERosendo: Small suggestion: `getItemsFromStorage()` returns undefined when there isn't a record for the current tab yet. In that case, accessing `tabStorage.caseId` below would throw.

We've seen users sometimes navigate directly to a docket report in a fresh tab, so it seems worth being a little defensive here. Defaulting to `{}` keeps that path safe and is consistent with the `tabStorage` && ... pattern used elsewhere (e.g. `utils.js:321`, `utils.js:341`)


```suggestion
    const tabStorage = (await getItemsFromStorage(this.tabId)) || {};
```

ERosendo (changes requested):
Thanks for putting this together! it's a nice improvement, and it should address the stale cache issue from https://github.com/freelawproject/courtlistener/issues/5247.

I left a few small suggestions, mostly around hardening and documentation, but I think the overall approach looks great.

src/pacer.js
playfulart: Adopted your version verbatim in 121ecfa — agreed it reads better top-to-bottom. Thanks!

src/pacer.js
playfulart: Good catch — it is a plurality, not a strict majority. Took your wording in 29e4871, and aligned the one other comment (in `handleDocketDisplayPage`) that had the same mistake. Also updated the PR description to match.

src/content_delegate.js
playfulart: You're right — I checked `handleAttachmentMenuPage` and it resolves its case id through `getPacerCaseIdFromPacerDocId` (the `docsToCases` map), so the attachment-menu claim was wrong. Took your comment in 8e8be8f.

src/content_delegate.js
playfulart: Applied in 9d02f4d, plus a regression spec for the fresh-tab path (storage yields `undefined`) — verified the spec fails without the guard. Suite is 188/188.

ERosendo (approved):

